PR #60, opened 2025-07-23
[ENH] Generalized PSeAAC

### LINKED ISSUE #59: [ENH] Generalized PSeAAC algorithm
The current implementation of the PSeAAC algorithm was designed solely to satisfy AptaNet, utilizing a version of the PSeAAC algorithm that AptaNet had specified.

The purpose of this PR is to generalize this algorithm, as it is a very popular method in bioinformatics. The [original paper](https://onlinelibrary.wiley.com/doi/10.1002/prot.1035) has over 1500 citations, and a [simple search on Google Scholar](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=pseudo+amino+acid+composition+aptamer&btnG=) will show numerous papers referencing it in the context of aptamer design.

The problem with the current implementation is:
1. It is built to give best results when k-mer length is 4-which is incorrect. A k-mer of length 4 gives the best results with 18 physicochemical properties, not 21 as used in the current implementation.
2. It uses 21 properties for all k-mer lengths, which does not lead to optimal results, as each different k-mer length has a preferred number of physicochemical properties.
3. Grouping properties into groups of 3 is constraining, and frankly too rigid, for absolutely no reason.
4. There is no allowance for grouping arbitrary indices together; it currently only allows groups of 3 in ascending order.

satvshr:
closes #59.

This PR moves the current PSeAAC algorithm into the aptanet directory and renames the old class to `AptaNetPSeAAC` (since it is aptanet specific).
This PR also makes the PSeAAC algorithm (in root) more flexible as:
1. You can choose how many physicochemical properties you want to use-it's no longer limited to 21 (enabled via `prop_indices`).
2. You can choose custom groups, which are no longer limited to size 3 (via `group_props`, which groups based on the order of elements in `prop_indices`).
3. You can define your own groups explicitly, instead of being limited to automatically constructed groups from `group_props` (via `custom_groups`).


fkiraly:
> I dont get why if default class calls match

I think a user is significantly more confused how to use the "generalized" class than the current class. This is definitely true at the current state of docstrings, which does not explain very well what the new variables would or could be, nor does it give any examples.
I am guessing that even in the case of perfect docstrings, the "generalized" class will be more confusing to the user, that may also be due to the chosen parametrization.

Also, it is easier to iterate on this class in review if it is a separate one, and we are less likely to reintroduce bugs in the existing class that is used elsewhere.

fkiraly (changes requested):
Thanks. Can you kindly put the generalized and the specific algorithm both in the `pseaac` folder?

satvshr:
> Thanks. Can you kindly put the generalized and the specific algorithm both in the `pseaac` folder?

Done.

fkiraly (changes requested):
I am not sure what you did, but now you seem to have two identical classes, one in `pseaac`, and one in `pseaac.aptanet`. The old class has been deleted.

Can you kindly do the following:

* restore the original class exactly as it was before this PR.
* add the new class to a module called `_features_general`, in `pseaac`.
* put tests in the same folder, `aptanet.tests`.

fkiraly:
> You mean `pseaac.tests`?

Yes, that is what I meant

> I would like to rename the old module to `_features_aptanet` then

Can you do the renaming in another PR? Just to avoid any accidents and keep the files tracked.

fkiraly:
> @fkiraly are you ok with this? I think we should have the generalized form and aptanet form be in separate directories, so that their feature sets can be different.

Not sure if we agree on the final state or what to do in this PR.

Regarding the end state:

* you say "separate directories", I think the algorithm should be in the same directory (`pseaac`)
* and in separate files, `_pseaac_aptanet` and `_pseaac_general`

In this PR specifically, I would recommend to avoid *renaming* any existing files to avoid merge conflicts later on. Renaming `_pseaac` tp `_pseaac` I would do in a separate PR.

satvshr:
> * and in separate files, `_pseaac_aptanet` and `_pseaac_general`

I disagree with that as stated above: "the properties which were used for aptanet will also then be used for generalized. And if someone wants to expand on the available property groups for generalized pseaac, aptanet pseaac will end up using the newly added ones too." nstead id rather have the aptanet version inside a sub-directory in `pseaac`.

Issue #55, opened 2025-07-19
[ENH] Add function to convert `pdb` to `String`.


satvshr:
Since AptaNet accepts `String` data format, for the example notebook we need a function to convert `pdb` input to `String`.

PR #61, opened 2025-07-23
[ENH] `Structure` to `String` helper function

satvshr:
Waiting for #62 to be merged.

Solves #55 by adding a helper function `struct_to_aaseq` in `utils`. 

Location of helper function is open to feedback given I have no clear idea where it should go. Just as a note, this uses `biopython` and will be a very important base function in the development of this library.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dev = [
     "ruff>=0.12.0",
     "pytest>=8.0.0",
     "pre-commit",
-    "biopython>=1.83"
 ]
 
 [tool.ruff]


fkiraly (approved):
quice simple - we should think about whether we need even a separate function for htis, but I suppose ok for now.

satvshr:
> quite simple - we should think about whether we need even a separate function for htis, but I suppose ok for now.

I mean, the function is something that will probably be used quite a bit, and in my opinion as long as it reduces any n number of lines of code to 1 function call, it is worthy of being a separate function.

PR #30, opened 2025-07-07
[ENH] AptaNet algorithm

satvshr:
Merge after #28.
Solves #13.
Adds AptaNet, a binary classification algorithm to predict if an aptamer will bind to the protein or not.

PR #68, opened 2025-07-31
[DOC] example notebook for AptaNet

### LINKED ISSUE #54: [DOC] Add an example notebook for AptaNet
Add an example notebook to show the entire pipeline from data to prediction for AptaNet.

satvshr:
Requires #61, #67, and #30.
closes #54


pyaptamer/datasets/_loaders/one_gnh.py
```diff
@@ -0,0 +1,25 @@
+import os
```
fkiraly, line 1: can you add this file in a separate PR please?

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dev = [
     "ruff>=0.12.0",
     "pytest>=8.0.0",
     "pre-commit",
+    "biopython>=1.83"
```
fkiraly, line 31: why do we require biopython?

fkiraly, line 31: duplicate

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
 dev = [
     "ruff>=0.12.0",
     "pytest>=8.0.0",
-    "pre-commit",
+    "pre-commit"
```
fkiraly, line 30: leave this please

fkiraly (changes requested):
This is still jumbled, I do not think my request was addressed well.

* first, explain only the minimal all-in-one. Do not give information on how to customize until you have explained the basic workflow.
    * also, don´t just construct the pipeline. Run it as well, and provide the full vignette!
* do not introduce imbalanced-learn or any other extranous tools until you have introduced the package internal customization
* split this into three parts. Only cover information necessary for part x in part x.

Does this make sense? Please ask if it is not clear what I am asking.

satvshr:
> * first, explain only the minimal all-in-one. Do not give information on how to customize until you have explained the basic workflow.

Is that not being done under "Build your own pipeline"?

> also, don´t just construct the pipeline. Run it as well, and provide the full vignette!

Both the pipelines are being constructed using the `pipeline` variable, since training and testing are common for both, I just put "running" it under the "Model training and prediction" section.

> do not introduce imbalanced-learn or any other extranous tools until you have introduced the package internal customization

 I can move the imblearn imports after I define the components `feature_transformer`, `net`, and `selector`, if that is what you want? But I cannot remove it.

> * split this into three parts. Only cover information necessary for part x in part x.

I assume these are the 3 parts you are referring to:
1. `AptaNetPipeline` without `NeighbourhoodCleaningRule`.
2. `AptaNetPipeline` with `NeighbourhoodCleaningRule`.
3. Custom-built pipeline with `NeighbourhoodCleaningRule`.

If this is what you want, point 2 is not possible because of [this](https://github.com/gc-os-ai/pyaptamer/pull/68?notification_referrer_id=NT_kwDOBrX53rUxNzkyNzY2MDQ0MjoxMTI1ODkyNzg#pullrequestreview-3135090329)

satvshr:
> Besides, is there not a pipeline that does not have the restrictions, e.g., from ImbalancedLearn?

The pipeline I use in the notebook for the third workflow is the one from imblearn.

> I see - then at least we should have an optional `neighborhood_cleaning=True` or similar flag in the AptaNet pipeline. All the work was about making the algorithm to be one-line-executable.

I think we had discussed this and chosen to not keep it as a part of the pipeline, as it depends on the training data.

satvshr:
> Either way, the high-level idea was to have the entire AptaNet pipeline as a one-liner. So not using the one-line specification and manually calling the preprocessors defeats that purpose.
> 
> I would not consider AptaNet complete until we have that for its common components. If the topic of neighborhood cleaning would have been brought up earlier, I would not have considered the AptaNet PR as complete.

If you want, I can open a separate issue to add the `NeighbourhoodCleaningRule` to `AptaNet`, but I will have to use the `imblearn` pipeline instead of the `sklearn` one. I personally do not think we need to add this as it depends on the imbalance of the training data, and does not seem that integral to the algorithm (plus an additional dependency).

Issue #77, opened 2025-08-04
[ENH] Add custom online loader for datasets


satvshr:
Merge #61 before this.

As discussed [here](https://github.com/gc-os-ai/pyaptamer/issues/64#issuecomment-3148267589), add a loader to take datasets from the `Protein Data Bank` and convert the `pdb` file provided to `String`, this will be done in 2 steps:
1. A `utils` function to convert `pdb` to `Structure`.
2. #61 to convert `Structure` to `String`.

PR #78, opened 2025-08-04
[ENH] Add a helper function to convert `pdb` to `Structure` and add a loader to use datasets from the Protein Data Bank

satvshr:
Solves #77.

### DIFF
--- a/pyaptamer/datasets/_loaders/_online_databank.py
+++ b/pyaptamer/datasets/_loaders/_online_databank.py
@@ -3,7 +3,7 @@ from Bio.PDB import PDBList
 from pyaptamer.utils.pdb_to_struct import pdb_to_struct
 
 
-def download_and_extract_sequences(pdb_id, overwrite=False):
+def load_from_rcsb(pdb_id, overwrite=False):
     """
     Download a PDB file from the RCSB Protein Data Bank and parse it into a `Structure`.
     Files are created in the current working directory.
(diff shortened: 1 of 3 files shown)


pyaptamer/datasets/_loaders/_online_databank.py
```diff
@@ -0,0 +1,27 @@
+from Bio.PDB import PDBList
+
+from pyaptamer.utils.pdb_to_struct import pdb_to_struct
+from pyaptamer.utils.struct_to_aaseq import struct_to_aaseq
+
+
+def download_and_extract_sequences(pdb_id):
+    """
```
fkiraly, line 8: this docstring is incomplete. Where is it downloading from, is it creating files on the hard drive, if yes where, etc

Also, why do we need this composition?

I would rmeove `struct_to_aaseq`, since only the AptaNet algorithm needs this.

I would design it without, and also enable caching - i.e., if the file already exists, and `force_download=False`, it does not get overwritten.

satvshr:
> Thanks - I would remove the string extraction from the loader, since without that it is generally usable (internet-to-in-memory) - without, it is specific only to AptaNet.

Yeah I think I remember bringing this up in the original issue linked to this PR but I had no idea what were the inputs required by AptaTrans and MAWS, will do what you mentioned above.

fkiraly (approved):

Issue #88, opened 2025-08-07
[BUG] docstring example failures in MCTS and PSeAAC


fkiraly:
@NennoMP, @satvshr, docstring examples are failing in MCTS and PSeAAC, could you kindly have a look?
https://github.com/gc-os-ai/pyaptamer/pull/83

NennoMP:
For `MCTS`, see #89. 
Most problems have been addressed but I have one issue which I described in the PR.

siddharth7113:
@NennoMP  assuming this was fixed for MCTS right?

NennoMP:
> [@NennoMP](https://github.com/NennoMP) assuming this was fixed for MCTS right?

Yes

Issue #48, opened 2025-07-15
AptaTrans: deep NN pipeline, layers


fkiraly:
We should refactor the AptaTrans deep NN pipeline to `torch` idiomatic components, and add it to the repository.

In this issue, we do not test or find out how it gets trained, just make sure that the components are present, the pipeline is complete, and there are basic run tests with random weights etc, in the repository.

PR #89, opened 2025-08-07
[ENH] Fix example failures in MCTS, Fix tests structure inconsistencies

NennoMP:
This PR is related to #88 but only resolves the failures related to `MCTS`. I should have fixed the failing examples for `MCTS`. However, I have one issue.

The example will still fail because I didn't define `model`. The problem is that on main I don't have `AptaTrans` yet. @fkiraly Could we just let them fail until aptatrans gets merged and then I update the docstring? Otherwise I could create a temporary mock `torch.nn` model class which gives me random scores.

**EDIT**: just discovered `# doctest: +SKIP` exist. Could I use them to skip failing examples until what I mentioned above is sorted out?

I did a few additional changes that are not really relevant, but just to point them out:
* `Aptamer.reconstruct(...)` is no longer private as I now want to use it inside `MCTS` so we return to the user both the "original" candidate and the "reconstructed" one. Doesn't change functionality, it's just a feature;
* resolved the tests structure inconsistencies, so we don't have to wait for #63, which may take some time due to the copyright issues.


fkiraly:
> EDIT: just discovered # doctest: +SKIP exist. Could I use them to skip failing examples until what I mentioned above is sorted out?

Sure, but it is easy to then forget about it...

May I suggest stacking on #48?

fkiraly:
fixed a few additional occurrences and added `doctest: +SKIP` to the instances that cannot be resolved at current state of the repo

Issue #82, opened 2025-08-05
[ENH] clean implementation of `DeepAptamer`


fkiraly:
After prioritization, we decided that `DeepAptamer` would be a nice next algorithm to "clean up" for easy use.

[code](https://github.com/YangX-BIDD/DeepAptamer) [paper](https://pubmed.ncbi.nlm.nih.gov/39897584/) [pdf](https://www.cell.com/molecular-therapy-family/nucleic-acids/pdf/S2162-2531(24)00323-8.pdf)

satvshr:
Blocker: `DeepAptamer` uses 4 shape properties which they got from a web server called `DNAshape`, there is an R repository by some authors of that paper called [`DNAshapeR`](https://github.com/TsuPeiChiu/DNAshapeR), there is no python repository for the same. My question is: Is there any way to run R code in python?

satvshr:
@KubiczekD @JaBirke In the paper (page 6) under "Pre-trained DeepAptamer could identify key nucleotides
essential for target binding" they mention VAE (which I think suggests where the aptamer can bind with the target protein) as a use-case. I remember Dennis talking about the use of DeepAptamer for this, but I would like you to know VAE is not a part of the architecture but the output from DeepAptamer is something VAE uses (There is no code on how the VAE is constructed either), just to avoid misunderstandings! Kindly correct me if I am wrong.

satvshr:
* The [paper](https://www.cell.com/molecular-therapy-family/nucleic-acids/pdf/S2162-2531(24)00323-8.pdf) mentions using a 2D convolutional filter in page 4 under "The integrated DeepAptamer model demonstrated superior performance compared with the individual CNN or BiLSTM model" but the [image](https://github.com/YangX-BIDD/DeepAptamer/blob/84ee416702630a4e244c78f92e673e0ddbcda28b/model/BCMA/cnn_bilstm/model.png) mentions using a 1D convolutional model instead.
* The image linked above also includes attention layers for some reason and when you try searching for the word attention in the official paper, it only links to 2 citations in references.

fkiraly:
> DeepAptamer uses 4 shape properties which they got from a web server called DNAshape

What are these as objects? Type, size, etc

satvshr:
> I would read this as "variational auto-encoder", this is a type of generative neural network

I was not doubting what it is, I am confused whether it is a part of DeepAptamer, but I am fairly certain it is not.

[...]

PR #98, opened 2025-08-14
[ENH] The `DeepAptamer` algorithm

satvshr:
solves #82

### DIFF
--- a/pyaptamer/deepatamer/model.py
+++ b/pyaptamer/deepatamer/model.py
@@ -1,0 +1,65 @@
+import torch
+import torch.nn as nn
+
+
+class DeepAptamer(nn.Module):
+    def __init__(self, dropout=0.1, lstm_hidden=100):
+        super().__init__()
+
+        # Sequence branch (input: B x 35 x 4) model1
+        self.seq_conv = nn.Conv1d(in_channels=4, out_channels=12, kernel_size=1)
+        self.seq_dense = nn.Sequential(
+            nn.Linear(12, 32),
+            nn.ReLU(),
+            nn.Linear(32, 4),  # no activation, passthrough features
+            nn.ReLU(),
+        )
+
+        # Shape branch (input: B x 126 x 1) model2
+        self.shape_conv = nn.Conv1d(in_channels=1, out_channels=1, kernel_size=100)
+        self.shape_pool = nn.MaxPool1d(kernel_size=20, stride=20)  # 27 -> 1
+        self.shape_dense = nn.Sequential(nn.Linear(1, 4), nn.ReLU())
+
+        # BiLSTM over concatenated features (36 x 4)
+        self.bi1 = nn.LSTM(
+            input_size=4, hidden_size=100, batch_first=True, bidirectional=True
+        )
+        self.dropout1 = nn.Dropout(dropout)
+
+        self.bi2 = nn.LSTM(
+            input_size=200,
+            hidden_size=lstm_hidden,
+            batch_first=True,
+            bidirectional=True,
+        )
+        self.dropout2 = nn.Dropout(dropout)
+
+        # Final classification head
+        self.head = nn.Linear(2 * lstm_hidden, 2)  # logits
+
+    def forward(self, x_seq, x_shape):
+        # Sequence branch
+        s = x_seq.permute(0, 2, 1)  # B x 4 x 35
+        s = self.seq_conv(s)  # B x 12 x 35
+        s = s.permute(0, 2, 1)  # B x 35 x 12
+        s = self.seq_dense(s)  # B x 35 x 4
+
+        # Shape branch
+        h = x_shape.permute(0, 2, 1)  # B x 1 x 126
+        h = self.shape_conv(h)  # B x 1 x 27
+        h = self.shape_pool(h)  # B x 1 x 1
+        h = h.permute(0, 2, 1)  # B x 1 x 1
+        h = self.shape_dense(h)  # B x 1 x 4
+
+        # Concatenate along sequence length axis: (35+1) = 36
+        x = torch.cat([s, h], dim=1)  # B x 36 x 4
+
+        # BiLSTM stack
+        x, _ = self.bi1(x)  # B x 36 x 200
+        x = self.dropout1(x)
+        x, (hn, _) = self.bi2(x)  # hn: (2, B, hidden)
+        x = torch.cat([hn[-2], hn[-1]], dim=1)  # B x (2*lstm_hidden)
+        x = self.dropout2(x)
+
+        logits = self.head(x)  # B x 2 (logits)
+        return logits
--- a/pyaptamer/deepatamer/model.py
+++ b/pyaptamer/deepatamer/model.py
@@ -7,18 +7,22 @@ class DeepAptamer(nn.Module):
         super().__init__()
 
         # Sequence branch (input: B x 35 x 4) model1
-        self.seq_conv = nn.Conv1d(in_channels=4, out_channels=12, kernel_size=1)
-        self.seq_dense = nn.Sequential(
+        self.model1 = nn.Sequential(
+            nn.Conv1d(in_channels=4, out_channels=12, kernel_size=1),
+            nn.MaxPool1d(kernel_size=1, stride=1),
             nn.Linear(12, 32),
             nn.ReLU(),
-            nn.Linear(32, 4),  # no activation, passthrough features
-            nn.ReLU(),
+            nn.Linear(32, 4),
+            nn.Softmax(),
         )
 
         # Shape branch (input: B x 126 x 1) model2
-        self.shape_conv = nn.Conv1d(in_channels=1, out_channels=1, kernel_size=100)
-        self.shape_pool = nn.MaxPool1d(kernel_size=20, stride=20)  # 27 -> 1
-        self.shape_dense = nn.Sequential(nn.Linear(1, 4), nn.ReLU())
+        self.model2 = nn.Sequential(
+            nn.Conv1d(in_channels=1, out_channels=1, kernel_size=100),
+            nn.MaxPool1d(kernel_size=20, stride=20),
+            nn.Linear(1, 4),
+            nn.ReLU(),
+        )
 
         # BiLSTM over concatenated features (36 x 4)
         self.bi1 = nn.LSTM(
--- a/pyaptamer/utils/_deepaptamer_utils.py
+++ b/pyaptamer/utils/_deepaptamer_utils.py
@@ -1,0 +1,66 @@
+import numpy as np
+
+
+def ohe(X):
+    """
+    One-hot encodes a batch of DNA sequences.
+
+    Each sequence is converted into a matrix of one-hot vectors.
+    Unknown characters are encoded as [0, 0, 0, 0].
+
+    Parameters
+    ----------
+    X : np.ndarray or str
+        - If str: a single DNA sequence.
+        - If np.ndarray: a 1D array of DNA sequences (strings), shape (n_sequences,).
+
+    Returns
+    -------
+    np.ndarray
+        A 3D NumPy array of shape (n_sequences, seq_len, 4),
+        where each sequence is one-hot encoded.
+        Column order is [A, T, C, G].
+    """
+    alphabet = "ATCG"
+    mapping = {base: i for i, base in enumerate(alphabet)}
+
+    if isinstance(X, str):
+        X = np.array([X])
+
+    n_seqs = len(X)
+    seq_len = len(X[0])
+
+    ohe_batch = np.zeros((n_seqs, seq_len, 4), dtype=int)
+
+    for i, seq in enumerate(X):
+        for j, base in enumerate(seq):
+            idx = mapping.get(base)
+            if idx is not None:
+                ohe_batch[i, j, idx] = 1
+
+    return ohe_batch
+
+
+def pad_sequences(X):
+    """
+    Pads DNA sequences to length 35 using 'N'. Raises an error if any sequence is longer
+    than 35.
+
+    Parameters
+    ----------
+    X : list of str or np.ndarray
+        List or array of DNA sequences (strings) of length ≤ 35.
+
+    Returns
+    -------
+    np.ndarray
+        A NumPy array of shape (n_sequences,) with sequences padded to exactly
+        35 characters.
+    """
+    X = np.asarray(X, dtype=str)
+    for seq in X:
+        if len(seq) > 35:
+            raise ValueError(f"Sequence length {len(seq)} exceeds 35: '{seq}'")
+
+    padded = np.array([seq.ljust(35, "N") for seq in X])
+    return padded
--- a/pyaptamer/deepatamer/__init__.py
+++ b/pyaptamer/deepatamer/__init__.py
@@ -1,0 +1,1 @@
+"""The deepatamer algorithm for binary binding affinity prediction."""
--- a/pyaptamer/deepatamer/preprocessing.py
+++ b/pyaptamer/deepatamer/preprocessing.py
@@ -1,0 +1,58 @@
+import numpy as np
+
+from pyaptamer.utils._deepaptamer_utils import ohe, pad_sequences
+
+
+def preprocess_x_ohe(X):
+    """
+    Preprocesses input DNA sequences for DeepAptamer.
+
+    This function pads each input sequence to length 35 using 'N' and one-hot encodes
+    them. The resulting array has shape (n_sequences, 35, 4), where each base is
+    encoded as:
+    - A → [1, 0, 0, 0]
+    - T → [0, 1, 0, 0]
+    - C → [0, 0, 1, 0]
+    - G → [0, 0, 0, 1]
+    - N or unknown → [0, 0, 0, 0]
+
+    Parameters
+    ----------
+    X : list of str or np.ndarray
+        A list or array of DNA sequences (strings), each of length ≤ 35.
+
+    Returns
+    -------
+    np.ndarray
+        A NumPy array of shape (n_sequences, 35, 4) representing one-hot
+        encoded sequences.
+    """
+    X_pad = pad_sequences(X)
+    X_ohe = ohe(X_pad)
+    return X_ohe
+
+
+def preprocess_x_pad(X):
+    # pad x with 2 zeros on both sides
+    # X_pad = 2 * ["0"] + X + 2 * ["0"]
+    return X
+
+
+def preprocess_y(y):
+    """
+    Preprocess labels into one-hot vectors.
+
+    Converts:
+        1 -> [1, 0]  (binder)
+        0 -> [0, 1]  (non-binder)
+
+    Args:
+        y (array-like): list or numpy array of binary labels (0 or 1)
+
+    Returns:
+        np.ndarray: one-hot encoded labels with shape (n_samples, 2)
+    """
+    one_hot = np.zeros((len(y), 2), dtype=int)
+    one_hot[y == 1] = [1, 0]
+    one_hot[y == 0] = [0, 1]
+    return one_hot
--- a/pyaptamer/deepatamer/tests/__init__.py
+++ b/pyaptamer/deepatamer/tests/__init__.py
@@ -1,0 +1,1 @@
+"""Tests for the deepatamer algorithm."""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ exclude = ["__pycache__", ".venv", "build", "dist", ".git"]
 
 [tool.ruff.lint]
 select = ["E", "F", "B", "I", "UP", "N", "C4"]
-ignore = ["N803", "N806"]
+ignore = ["N803", "N806", "N812"]
 
 [tool.ruff.format]
 quote-style = "double"
--- a/x.py
+++ b/x.py
@@ -1,0 +1,24 @@
+import numpy as np
+
+
+def preprocess_y(y):
+    """
+    Preprocess labels into one-hot vectors.
+
+    Converts:
+        1 -> [1, 0]  (binder)
+        0 -> [0, 1]  (non-binder)
+
+    Args:
+        y (array-like): numpy array of binary labels (0 or 1)
+
+    Returns:
+        np.ndarray: one-hot encoded labels with shape (n_samples, 2)
+    """
+    one_hot = np.zeros((len(y), 2), dtype=int)
+    one_hot[y == 1] = [1, 0]
+    one_hot[y == 0] = [0, 1]
+    return one_hot
+
+
+print(preprocess_y(np.array([1, 0, 1, 0])))  # Example usage


satvshr:
@NennoMP Could you checkout the [model architecture](https://github.com/gc-os-ai/pyaptamer/blob/659e4d00d9a5a437fa6d30c7c90b1f59dbba9e6b/pyaptamer/deepatamer/model.py#L6)? It should match their implementation [here](https://github.com/YangX-BIDD/DeepAptamer/blob/84ee416702630a4e244c78f92e673e0ddbcda28b/script/model.py#L240), which I think it is doing. Kindly ignore the comments left by GPT, I made it put them there in order to understand the flow of shapes throughout the network. I will remove them later.

NennoMP:
Sure. I will do it during the weekend.

pyaptamer/deepatamer/model.py
```diff
@@ -0,0 +1,65 @@
...
+
+class DeepAptamer(nn.Module):
+    def __init__(self):
+        super().__init__()
+
+        # Sequence branch (B, 35, 4) -> (B, 35, 4)
+        self.seq_conv = nn.Conv1d(in_channels=4, out_channels=12, kernel_size=1)
+        self.seq_fc = nn.Sequential(
+            nn.Linear(12, 32), nn.ReLU(), nn.Linear(32, 4), nn.Softmax(dim=-1)
+        )
```
NennoMP, line 14: This looks like it should implement the first code block in `create_combine(...)`. I think this is missing a `MaxPooling1D`. Also, in the forward pass there is not one via `torch.nn.functional`.

pyaptamer/deepatamer/model.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Shape branch (B, 126, 1) -> (B, 1, 4)
+        self.shape_conv = nn.Conv1d(in_channels=1, out_channels=1, kernel_size=100)
+        self.shape_fc = nn.Sequential(nn.Linear(1, 4), nn.ReLU())
```
NennoMP, line 18: This should be `model2` in `create_combine(...)`. I see in this case, the `MaxPooling1D` is in the forward pass. This seems ok, just two tips:

1. I would suggest to define all components at initialization time visa `self` attributes, rather than having some implemented via `torch.nn.functional`. The main reason in the first case, if you print the model you have a comprehensive overview of all its components. However, components applied on-the-fly `torch.nn.functional` are not printed and you may not know they are there without looking at the code.

2. For max pooling, if you want the the stride to be the same as the kernel size, you can simply do `MaxPool1D(kernel_size=100)`. The stride is, by default, initialized as `stride = kernel_size`.

pyaptamer/deepatamer/model.py
```diff
@@ -0,0 +1,65 @@
...
+        # BiLSTM stack (matching TF's 100 units per direction)
+        self.bi1 = nn.LSTM(
+            input_size=4, hidden_size=100, batch_first=True, bidirectional=True
+        )
+        self.dropout1 = nn.Dropout(0.1)
+
+        self.bi2 = nn.LSTM(
+            input_size=200, hidden_size=100, batch_first=True, bidirectional=True
+        )
+        self.dropout2 = nn.Dropout(0.1)
```
NennoMP, line 29: I think you can simplify this by setting `nn.LSTM` parameter `num_layers = 2`. As far as I know it should stack two LSTMs on top of each other and apply dropout in-between.
What it does not do, is apply dropout after the last layer, so that's one thing you should apply manually.

```python
self.bilstm = nn.LSTM(
    input_size=4, hidden_size=100, num_layers=2, 
    batch_first=True, bidirectional=True, dropout=0.1,
)
self.dropout = Dropou(0.1)
```

That said, isn't a fuly-connected head + sigmoid missing here? In the authors' code they have one right-after the two BiLSTMs, still in the context of `model3`.

NennoMP (commented):
I left a few comments.

I haven't finished looking at the authors' code. To be honest, it has to be one of the worst, monolithic code bases I have seen 😁.

EDIT: I just noticed is that some shapes/layer dimension in their code do not match the paper? In Figure 3 the output of the BiLSTMs is highlighted as `(m, 124)`, which means these layers have a `hidden_size = 64` which is then multiplied by factor `2` due to the bidirectionality. In their code they use `hidden_size = 100` in `create_combine(...)`. They do use `hidden_size = 64` in `create_bilstm(...)` though.

I am not sure which methods they actually use to build the model or not. If you discover more let me know.

satvshr:
Cannot use lightning due to lower numpy version requirements because of `Deep DNAshape`

satvshr:
@fkiraly How to setup an environment to run tests for a downgraded version of `numpy` and `tensorflow` (`deepDNAshape`)

satvshr:
@fkiraly Tests passing locally in my own environment, parking this PR till the `DeepDNAshape` problem gets sorted out.

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,81 @@
+__author__ = "satvshr"
+__all__ = ["DeepAptamerNN"]
+
+import torch
+import torch.nn as nn
+
+
+class DeepAptamerNN(nn.Module):
+    def __init__(self, optimizer=None):
+        super().__init__()
```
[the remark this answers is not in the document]
satvshr: Makes sense, will do

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Sequence branch (B, 35, 4)
+        self.seq_conv = nn.Conv1d(in_channels=4, out_channels=12, kernel_size=1)
+        self.seq_fc = nn.Sequential(
+            nn.Linear(12, 32),
+            nn.ReLU(),
+            nn.Linear(32, 4),
+            nn.Softmax(dim=-1),
+        )
```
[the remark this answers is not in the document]
satvshr: Since it's not a part of the model architecture given it does nothing, id say let the model components be the same at least while still being modifiable? Like if we introduce it back we will be changing the architecture by introducing a new  component? Let me know if that sounds right!

pyaptamer/deepatamer/_pipeline.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Methods
+    -------
+    predict(seqs)
+        Compute ranked binding affinity scores for one or more DNA
+        sequences. Returns a list of dictionaries with each sequence
+        and its predicted binding probability.
+    """
+
+    def __init__(self, model, use_126_shape=True, device="cpu"):
```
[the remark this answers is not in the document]
satvshr: I completely agree, was not a fan of the variable name either and was honestly waiting for franz to suggest a name 😅 I'll change it to what you're suggesting as it is way better!

pyaptamer/deepatamer/_model.py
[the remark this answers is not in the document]
NennoMP: > Since it's not a part of the model architecture given it does nothing, id say let the model components be the same at least while still being modifiable? Like if we introduce it back we will be changing the architecture by introducing a new component? Let me know if that sounds right!

You are not wrong... I suggested re-adding it because I think it could be awkward for someone that has read the paper and sees our implementation. 
In the paper they (erroneously) mention max pooling in both branches, but then in their code one of them does nothing due to kernel size.

To be honest there is no wrong-right approach here, I would ask @fkiraly what he thinks. I tried to check whether the paper discusses kernel sizes but nope...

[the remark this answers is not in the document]
fkiraly: personally, I would leave it in with a default of 1 but customizable params, and give an explanation in the docstring?

fkiraly:
> @fkiraly Any idea how to setup an environment to run tests for a downgraded version of `numpy` and `tensorflow` (`deepDNAshape`)?

locally, you install your environment with a `requirements.txt` or version bound specifiers when calling `pip install` or equivalent. On remote, you would have to modify `pyproject.toml`, but I would rather not put a `numpy<1.24` bound in there.

pyaptamer/deepatamer/_pipeline.py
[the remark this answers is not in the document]
satvshr: I mentioned this today in the standup, I also mentioned it over [here](https://github.com/gc-os-ai/pyaptamer/issues/82#issuecomment-3193622643).

TLDR: Both python(DeepDNAshape) and R(original) implementations of DNAshape output 138 vectors, but in the case of R it contains 12 `NA` vectors (hence 138-12 = 126). `DeepAptamer` uses 126 vectors after removing the `NA` values from the R implementation.

pyaptamer/deepatamer/_model.py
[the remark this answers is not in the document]
satvshr: Done

satvshr:
@NennoMP can you have a look and resolve the comments made above once you figure they have been resolved?

pyaptamer/deepatamer/_model.py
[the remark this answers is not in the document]
NennoMP: Looks good. I would make a minor adjustment:
* just one `dropout` parameter for both dropouts, to make the `__init__` method a little bit slimmer. They use the same value anyway in their code so I wouldn't bother having two distinct ones.

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,128 @@
...
+        x = self.dropout(x)
+
+        x = x.unsqueeze(1)
+        x, _ = self.attn(x, x, x)
+
+        x = x.squeeze(1)
+
+        return self.head(x)
+
+    def train_loop(self, X_ohe, X_shape, y, epochs=10, device="cpu"):
```
NennoMP, line 115: Do we really want training code inside the model class? I think that this doesn't go well with both `pytorch` and `lightning`, independently of what we end up using to train `DeepAptamer`.

For `pytorch` we would want the training code in a separate class or in the notebook itself (if not too complex). For `lightning` we won't have training code at all but just calls to its `Trainer` class. If we are keeping this as a temporary method for testing purposes ok, but mark it for removal so we don't forget about it.

pyaptamer/deepatamer/_pipeline.py
```diff
@@ -0,0 +1,111 @@
+__author__ = "satvshr"
+__all__ = ["DeepAptamerPipeline"]
+
+import numpy as np
+import torch
+
+from pyaptamer.deepatamer._preprocessing import preprocess_seq_ohe, preprocess_seq_shape
+
+
+class DeepAptamerPipeline:
```
NennoMP, line 10: I imagine you're still working on docstrings but since this seems complete I will provide feedback:
* in `numpydoc` in-text references are cited as `[1]_`;
* `References` usually goes after `Parameters` and `Methods` but before `Examples`
* Let's remember to cite [2]_ and [3]_, which are currently specified but not cited.

NennoMP (commented):
Resolves some discussions and added new comments.

The architecture's class is definitely better now, I only provided feedback on minor things and on the training method which I think should go elsewhere later one.

Docstrings are probably still work-in-progress so didn't look at them too much.

pyaptamer/deepatamer/_model.py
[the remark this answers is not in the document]
satvshr: > They use the same value anyway in their code so I wouldn't bother having two distinct ones.

Thought about it when I was doing it too and decided to keep it as there is nothing wrong with making it customizable too?

pyaptamer/deepatamer/_model.py
NennoMP: > `lightning` requires a [`training_step`](https://lightning.ai/docs/pytorch/stable/starter/converting.html#configure-training-logic)to be defined in the class, but not the entire logic. I will move this to the notebook itself as it is not complex?

But this class wouldn't work with `lightning` anyway no? Because we are inheriting from `nn.Module` rather than from `LightningModule`. 

I think we decided to go in the direction of having models be compatible with both `pytorch` and `lightning`? (correct me if I am wrong, I genuinely don't remember). If so, definitely remove it from here and put it in the model class that will act as a wrapper for `lightning`. Otherwise this class needs to inherit `LightningModule`.

pyaptamer/deepatamer/_model.py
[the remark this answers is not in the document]
satvshr: Changed it

pyaptamer/deepatamer/_pipeline.py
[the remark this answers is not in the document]
NennoMP: > Can this be resolved or is there something else to work on @NennoMP ?

I left it open because Franz was also interested in this and I don't know what he thinks about the new version. For me it's ok, but I would leave it open for his reference.

pyaptamer/deepatamer/_model.py
NennoMP: > > But this class wouldn't work with lightning anyway no?
> 
> Nope, due to lower package requirements (in this case numpy) `deepDNAshape` and `lightning` are not compatible. Checkout #112. I was just answering your question, could you answer this:
> 
> > I will move this to the notebook itself as it is not complex?

Then yes, you never see training code inside `nn.Module` instances. I would say, for now, put it in the notebook as it is very simple. Then, later we can think about making it compatible with the `Solver` class I created to AptaTrans. To make it work we would need to have data in pytorch dataloaders for instance. I wouldn't worry about this now though.

pyaptamer/deepatamer/_pipeline.py
NennoMP: > > Let's remember to cite [2]_ and [3]_, which are currently specified but not cited.
> 
> Why do I need to cite them?

Well if you put an entry in `References` you need to cite them somewhere, otherwise there is no point in having them. We know that [3]_ is the official repository code for `DeepAptamer` and [2]_ are some tools used in pre-processing or whatever. However, this may not be so clear to someone else.

Since [2]_ is jsut a package I wouldn't even cite it and remove the entry entirely. For [3]_, since it's just a reference to the original implementation I would pust a one-liner in the docstring: `Original implementation: https://github.com/YangX-BIDD/DeepAptamer`.

satvshr: Franz wanted them cited as far as I remember

pyaptamer/deepatamer/_model.py
satvshr: Done

pyaptamer/aptanet/_pipeline.py
```diff
@@ -13,7 +13,7 @@
 
 class AptaNetPipeline:
     """
-    AptaNet algorithm for aptamer–protein interaction prediction (Emami et al., 2021)
+    AptaNet algorithm for aptamer–protein interaction prediction ..[1]
```
NennoMP, line 16: Should be referenced as `[1]_` (see [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#references)).

pyaptamer/aptanet/_pipeline.py
```diff
@@ -13,7 +13,7 @@
...
 
     Implements the AptaNet algorithm, a deep learning method that combines
```
NennoMP, line 18: The docstring should incude the `Attributes` fields for parameter not initialized via `__init__` parameters. This is the case for `self.pipeline_`.

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,194 @@
...
+
+class DeepAptamerNN(nn.Module):
+    """
+    DeepAptamer neural network model for aptamer–protein interaction prediction.
+
+    This architecture integrates:
+
+    - A sequence branch using convolutional and fully-connected layers to
+      process one-hot encoded aptamer sequences.
+    - A structural (DNA shape) branch using convolution + pooling + dense layers.
```
NennoMP, line 16: I would add one sentence here to give more details about that the structural branch processes? What is a DNA shape?

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,194 @@
...
+    - A BiLSTM for capturing sequential dependencies.
+    - Multi-head self-attention for contextual feature refinement.
+    - A final classification head for binary binding prediction.
+
+    Parameters
+    ----------
+    seq_conv_in : int, optional
```
NennoMP, line 23: I would add default values for each parameter, e.g. ` seq_conv_in : int, optional, default=4` and so on.

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,194 @@
...
+        Number of BiLSTM layers. Default is 2.
+
+    dropout : float, optional
+        Dropout probability applied after the BiLSTM. Default is 0.1.
+
+    optimizer : torch.optim.Optimizer or None, optional
+        Optimizer for training. If None, defaults to Adam with lr=0.001.
+
+    Attributes
+    ----------
```
NennoMP, line 69: Missing `self.dropout` attribute.

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.attn = nn.MultiheadAttention(
+            embed_dim=2 * self.bilstm_hidden_size, num_heads=1, batch_first=True
+        )
+
+        self.head = nn.Linear(2 * bilstm_hidden_size, 2)
+
+        self.optimizer = self.optimizer or torch.optim.Adam(self.parameters(), lr=0.001)
+
+    def forward(self, x_ohe, x_shape):
```
NennoMP, line 172: I would add a docstring that specifies the shapes of inputs and outputs tensors/vectors. For an example you can see the forward passes of AptaTrans [here](https://github.com/gc-os-ai/pyaptamer/blob/main/pyaptamer/aptatrans/model.py).

pyaptamer/deepatamer/_pipeline.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    DeepAptamer algorithm for aptamer–protein interaction prediction _[1]
```
NennoMP, line 12: `[1]_`

pyaptamer/deepatamer/_pipeline.py
```diff
@@ -0,0 +1,116 @@
...
+        self.device = device
+
+    def predict(self, seqs):
+        """
+        Predict binding affinity scores for one or more sequences.
+
+        Parameters
+        ----------
+        seqs : str or list of str
+            DNA sequence(s), each length ≤ 35.
```
NennoMP, line 72: Does this mean that it works only for (aptamer) dna sequences that are no longer than `35`? Why so? Otherwise why do we expected them to be  of `length ≤ 35`. It seems restrictive...

pyaptamer/deepatamer/_preprocessing.py
```diff
@@ -0,0 +1,103 @@
...
+        A NumPy array of shape (35, 4) representing the one-hot
+        encoded sequence.
+    """
+    seq_pad = pad_sequence(seq)  # pads to 35
+    seq_ohe = ohe(seq_pad)  # one-hot encode (shape 35 × 4)
+    return seq_ohe
+
+
+# input will be of size (n_shapes(4), shape_vector_size)
+def preprocess_seq_shape(seq, use_126_shape=True):
```
NennoMP, line 43: I would rename `use_126_shape` variable as we did in the pipeline/model to `full_dna_shape`.

pyaptamer/utils/_deepaptamer_utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+    for j, base in enumerate(seq):
+        idx = mapping.get(base)
+        if idx is not None:
+            ohe_matrix[j, idx] = 1
+
+    return ohe_matrix
+
+
+def pad_sequence(seq):
```
NennoMP, line 41: I guess this is the reason why we expect aptamer sequences of at most `35` nucleotides right? Probably it was some hardcoded value in the authors' code.

Is there any reason why `DeepAptamer` wouldn't work if we have longer sequences? Because we could simply have a `max_len` parameter for this padding, to specify how long the sequences should be. In AptaTrans paddings works like this, which a user-specified maximum length.

pyaptamer/deepatamer/_preprocessing.py
```diff
@@ -0,0 +1,103 @@
...
+
+        norm_features.append(arr_norm)
+
+    # Step 2: Concatenate all features into one flat vector
+    seq_flat = np.concatenate(norm_features).reshape(1, -1)
+
+    return seq_flat
+
+
+def preprocess_y(y):
```
NennoMP, line 86: Why do we need to one-hot-encode already numerical labels?

Also, is this used somewhere? Not sure whether we need a method for this.

satvshr:
> Why do we need to one-hot encode already numerical labels (e.g., 1 to [1, 0]) in preprocess_y? Also, I don't see the method being used anywhere.

This is something I forgot to ask you and Franz! They do in the paper but I have no idea why.

> * Currently, `DeepAptamer` expects DNA sequences of length `<= 35` due to how padding is applied in `pad_sequence`. Can't we make this generic via a parameter? and pass a default value of `35` from the pipeline if that's that value used by the authors? I feel like otherwise this is very restrictive for the aptamer inputs.

Was thinking the same thing while talking about PSeAAC today, what was the conclusion on that topic by the way?

pyaptamer/deepatamer/_preprocessing.py
satvshr: Thought I made that change, will do

NennoMP:
> > Why do we need to one-hot encode already numerical labels (e.g., 1 to [1, 0]) in preprocess_y? Also, I don't see the method being used anywhere.
> 
> This is something I forgot to ask you and Franz! They do in the paper but I have no idea why. The method is not being used anywhere right now as it will be a part of the notebook.

I wouldn't bet on it, but I am pretty sure they do not need to do it. This is a multi-cass classification problem (binary, to be specific), and they probably confused it with a multi-output classification problem, where for each sample you need to predict `n` classes.

I would attempt to remove the one-hot-encoding on the labels and make it work with numerical labels. If something fails the cause is probably in their training code because they may expect one-hot-encoder labels.

> > * Currently, `DeepAptamer` expects DNA sequences of length `<= 35` due to how padding is applied in `pad_sequence`. Can't we make this generic via a parameter? and pass a default value of `35` from the pipeline if that's that value used by the authors? I feel like otherwise this is very restrictive for the aptamer inputs.
> 
> Was thinking the same thing while talking about PSeAAC today, what was the conclusion on that topic by the way?

The consensus is to make `PSeAAC` more generic and remove the constraint of protein sequence of length `>= 30`. Similarly, we should remove these constraints from `DeepAptamer` as well.

For PSeAAC, I will create a small PR to generalize expected length of protein sequences, I will check your PR #60 to see what you already did to generalize the algorithm.

satvshr:
> we should remove these constraints from DeepAptamer as well.

Already did that in the latest commit.

pyaptamer/deepatamer/_pipeline.py
```diff
@@ -0,0 +1,116 @@
...
+                "score": float (probability of binding, from [p_bind, p_not_bind])
+            }
+            Sorted from high to low by score.
+        """
+        if isinstance(seqs, str):
+            seqs = [seqs]
+
+        ohe_list, shape_list = [], []
+        for seq in seqs:
+            ohe_list.append(preprocess_seq_ohe(seq))
```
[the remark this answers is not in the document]
satvshr: That makes sense, will do that

satvshr:
Will addressthe  above comments in the thread, but the absolute features values like 4 and 126 can't be changed as they are fixed outputs, 4 because of ohe of the nucleotides into vectors of ACGT. 126 because that is the fixed shape output.

NennoMP:
> Will addressthe above comments in the thread, but the absolute features values like 4 and 126 can't be changed as they are fixed outputs, 4 because of ohe of the nucleotides into vectors of ACGT. 126 because that is the fixed shape output.

Ok, `4` being fixed makes sense but is it true also for `126`. It can be either `126` or `138` no? Depending on whether in the pipeline we use `full_dna_shape = False` or `full_dna_shape = True`. At least that's what the docstring in the pipeline says.

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,215 @@
...
+        self.seq_conv_linear_out = seq_conv_linear_out
+        self.shape_conv_kernel_size = shape_conv_kernel_size
+        self.shape_pool_kernel_size = shape_pool_kernel_size
+        self.shape_pool_stride = shape_pool_stride
+        self.bilstm_hidden_size = bilstm_hidden_size
+        self.bilstm_num_layers = bilstm_num_layers
+        self.dropout_val = dropout
+        self.optimizer = optimizer
+
+        # Sequence branch (B, seq_len, 4)
```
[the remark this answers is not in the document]
satvshr: > If you really want to keep them as comments, have them for all architectural components for consistency (e.g., also for `self.bilstm`, `self.head`, etc.).

It's just a preference; I don't think I need to change anything unless it conflicts with any design principle.

pyaptamer/deepatamer/_model.py
```diff
@@ -0,0 +1,215 @@
...
+            nn.MaxPool1d(
+                kernel_size=self.seq_pool_kernel_size, stride=self.seq_pool_stride
+            ),
+            nn.Linear(self.seq_conv_out, self.seq_linear_hidden_dim),
+            nn.ReLU(),
+            nn.Linear(self.seq_linear_hidden_dim, self.seq_conv_linear_out),
+            nn.Softmax(dim=-1),
+        )
+
+        # Shape branch (B, 1, 126)
```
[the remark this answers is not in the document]
satvshr: Changed to `(B, 1, 126/138)`

pyaptamer/deepatamer/_pipeline.py
```diff
@@ -0,0 +1,117 @@
...
+        for seq in seqs:
+            ohe_list.append(preprocess_seq_ohe(seq, seq_len=max_len))
+            shape_list.append(preprocess_seq_shape(seq))
+
+        X_ohe = torch.tensor(
+            np.array(ohe_list), dtype=torch.float32, device=self.device
+        )
+        X_shape = torch.tensor(
+            np.array(shape_list), dtype=torch.float32, device=self.device
+        )
```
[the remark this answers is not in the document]
NennoMP: By looking at `torch.tensor` [docs](https://docs.pytorch.org/docs/stable/generated/torch.tensor.html), its `device` parameter is expected to be a `torch.device` instance. This is also shown in the example:
```
cuda0 = torch.device('cuda:0')
torch.ones([2, 4], dtype=torch.float64, device=cuda0)
```

Not sure why it works when feeding it a string, but if the docs specify a a torch.device I would be consistent with it. Maybe this is not intended behavior and in a future version they will fix it.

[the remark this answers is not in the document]
satvshr: > This is also shown in the example:

Yeah I did see the example, what you are saying makes sense but I have been doing this ever since I have been using torch 😅 Yup, bit of researching into it and it is an accepted convention.
> The [torch.device](https://docs.pytorch.org/docs/stable/tensor_attributes.html#torch.device) argument in functions can generally be substituted with a string. This allows for fast prototyping of code.

# Example of a function that takes in a torch.device
cuda1 = torch.device('cuda:1')
torch.randn((2,3), device=cuda1)
# You can substitute the torch.device with a string
torch.randn((2,3), device='cuda:1')

[Source](https://docs.pytorch.org/docs/stable/tensor_attributes.html#torch.device) (under `Note`)

pyaptamer/deepatamer/_model.py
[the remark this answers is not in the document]
NennoMP: The design choice is fine, but I find it strange to have the input shape specified via a comment for only some components. Not really consistent no?

It would be a different story if there is a specific reason for specifying the shape for the sequence branch but not for, say, the `self.bilstm` or `self.attn`. Is there a specific reason?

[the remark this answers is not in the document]
satvshr: I am only doing what I am doing rn as they are 2 separate branches which concatenate in the lstm head, the 2 shape comments are the shapes of the inputs in the `forward` method. If it was all in one branch I would not comment on it but since its in 2 branches which connect later on (and it is the input), I have mentioned the shapes so that the user knows which vector is being processed by which part of the network.

pyaptamer/deepatamer/_preprocessing.py
satvshr: > Still need to adress this one. I don't remember if we already discussed this, maybe you want to bring it up during the weekly?

This is how it should be! Asked @fkiraly about it and he suggested looking into the torch API, and [BCELoss does expect 2d inputs](https://docs.pytorch.org/docs/stable/generated/torch.nn.BCELoss.html) (check example) !

satvshr:
@NennoMP can you resolve old convos if they have been fixed now?

fkiraly:
look at the tests, they are failing due to import errors

satvshr:
> I see, so this is the point where we are going to park this PR, right?

If it is in a "mergable" state, yes.

PR #113, opened 2025-08-24
Rename `_aptanet_nn.py` to `_model.py` for consistency

satvshr:


satvshr:
@fkiraly quick merge

fkiraly (changes requested):
I would actually not do this - the term "model" is not very clear as to what a file contains.

The user will also not interact with the module name directly, since it is private, so I think it is better to make it as descriptive as possible for developers who are unfamiliar with the module.

satvshr:
> The user will also not interact with the module name directly, since it is private, so I think it is better to make it as descriptive as possible for developers who are unfamiliar with the module.

do you suggest renaming the nn files to `_{model}_nn` then?

PR #117, opened 2025-08-25
Make AptaNet compatible with MCTS

### LINKED ISSUE #97: [ENH] AptaNet in-silico-selex or optimization
Looking through #68, it appears to me that AptaNet as current can be used only for interaction prediction, but not for aptamer design.

While this is identical with the scope of the paper, it is not how it would be used in the lab.

The natural extension is to combine a fitted AptaNet with a string input optimization algorithm such as MCTS, or some other form of in-silico selex.

Given the "incompleteness" compared to the aptamer design task, it feels like a higher priority than adding another algorithm?

In terms of a design, we should build an experiment that follows the same API as @NennoMP´s experiment, and takes as init arguments the protein `Structure`, and a trained AptaNet model?

FYI @satvshr, also @KubiczekD, @JaBirke - what do you think?

* priority of using Aptanet for aptamer design via in-silico selex vs new algorithm
* API choice of "experiment" to make it compatible with MCTS
* optional: new in-silico selex variants such as mutation based gradient free optimization (FYI @SimonBlanke)

Update: we decided to schedule this as a work item, to wrap AptaNet in an "Experiment" that predict binding affinity, and then use this in MCTS as a "predict aptamer for target" algorithm.

NennoMP:
Makes `AptaNet `compatibles with `MCTS`, mainly by:
* creating a dedicated aptamer evaluation experiments, called `AptamerEvalAptaNet`;
* add an option for aptanet pipeline `predict(...)` method to output the probability scores rather than the class labels (the raw probabilities are needed by MCTS).

This PR resolves #97.

Main modules to look at for review:
* `experiments._aptamer_eval.py`, specifically class `AptamerEvalAptaNet`;
* `predict(...)` method in `aptanet._pipeline.py` and `aptanet._feature_classifier.py`.


satvshr:
Your main branch does not seem updated, can you merge main so that I can see the changes made to AptaNet only?

Issue #34, opened 2025-07-08
AptaNet discrepancy between implementation and paper


satvshr:
TL;DR: My PR (#29) currently faces two issues: (1) the amino acid frequencies are [divided by 20 (number of possible amino acids (A...Y)) in the official implementation](https://github.com/nedaemami/AptaNet/blob/8f1f4d1fd13c97b84b828e517f7e1355366a7f07/target_encoding.py#L184) to calculate the `denominator_val`, which is used to calculate the feature vectors (instead of the target sequence length (normalization)), which seems to contradict [the paper]((https://www.nature.com/articles/s41598-021-85629-0.pdf)); and (2) even when using 20, the output vectors which we get at the end of pseaac don't match the original implementation: there are small but consistent discrepancies.

There are 2 main problems I would like to address with my PR:
1. In the official repository, they divide all amino acid frequencies by a fixed number, which is the total number of available amino acids that can exist, whereas it should be divided by the length of the target sequence instead, as mentioned in the paper in [page 14](https://www.nature.com/articles/s41598-021-85629-0.pdf) in the start of the second last para.
2. Even if I divide the frequency of the amino acids by 20, my final vectors do not match the ones being yielded by the official implementation in the end after the entire process is completed, and there are small discrepancies everywhere (we are comparing 2 vectors of length 350), with the error margin being `1e-3`. The format for each entry is `(<vector index>, <my value>, <official value>)`:
[...]

PR #126, opened 2025-08-28
Make PSeAAC more configurable

NennoMP:
Makes `PSeAAC` more configurable by enabling customization of `lambda` and `weight` parameters.

Main changes:
*  `lambda` and `weight` are now `__init__` parameters;
* updated a few docstrings here and there. Some were not consistent with `numpydoc` (e.g., missing attributes, methods after examples, etc.);
* moved in-method imports at the top of the module.

A few things to mention. The output of `transform(...)` being of length `20 + 30 = 50` was not a coincidence. In particular, the second part of length 30 depends on the value of lambda. I updated the docstring accordingly.

I noticed that `PSeAAC` has pretty bad test coverage, maybe due to #34 not being solved yet? I think we should open a separate issue and add a comprehensive test suite for the algorithm so we don't forget about it.

FYI @fkiraly @satvshr

### DIFF
--- a/pyaptamer/pseaac/_features.py
+++ b/pyaptamer/pseaac/_features.py
@@ -54,10 +54,10 @@ class PSeAAC:
 
     - 20 normalized amino acid composition features (frequency of each standard
     amino acid)
-    - 30 sequence-order correlation features based on physicochemical similarity between
-    residues.
-    These 50 features are computed for each of 7 predefined property groups,
-    resulting in a final vector of length 350 ((20 + 30) * 7 = 350).
+    - `self.lambda_val` sequence-order correlation features based on physicochemical
+    similarity between residues.
+    These (20 + `self.lambda_val`) features are computed for each of 7 predefined
+    property groups, resulting in a final vector of length (20 + `self.lambda_val`) * 7.
 
     See `transform` method for usage.
 
@@ -65,6 +65,8 @@ class PSeAAC:
     ----------
     lambda_val : int, optional, default=30
         The lambda parameter defining the number of sequence-order correlation factors.
+        This also determines the minimum length allowed for input protein sequences,
+        which should be of length greater than `lambda_val`.
     weight : float, optional, default=0.15
         The weight factor for the sequence-order correlation features.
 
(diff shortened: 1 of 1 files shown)


satvshr:
> I think we should open a separate issue

I'd prefer if you just mentioned it in the original issue to broaden the scope of the issue instead of creating a new one and making it more cluttered than it needs to be, also kindly mention which tests youd rather add too.

satvshr:
LGTM @fkiraly

satvshr:
> Could we stop skipping the test or do we strictly need to sort out #34 before that?

Ah you are talking about the main test! Yes #34 needs to be sorted out before commenting it out. If that test is removed I will ensure the pipeline runs with correct input and fails for wrong input, you can ignore it for now but I will definitely ping you when I work on PSeAAC again and make a PR for #34.

pyaptamer/pseaac/_features.py
```diff
@@ -57,32 +59,48 @@ class PSeAAC:
     These 50 features are computed for each of 7 predefined property groups,
     resulting in a final vector of length 350 ((20 + 30) * 7 = 350).
 
+    See `transform` method for usage.
+
+    Parameters
+    ----------
+    lambda_val : int, optional, default=30
+        The lambda parameter defining the number of sequence-order correlation factors.
```
fkiraly, line 67: it should be said that sequence lengths should be at least `lambda_val`

pyaptamer/pseaac/_features.py
```diff
@@ -187,55 +200,47 @@ def transform(self, protein_sequence):
...
-            A 1D NumPy array of length 50 * number of normalized physiochemical (NP)
-            property groups of amino acids (7).
-            Each 50-element block consists of:
+            A 1D NumPy array of length (20 + `self.lambda_val) * number of normalized
+            physiochemical (NP) property groups of amino acids (7).
+            Each element consists of:
             - 20 normalized amino acid composition features
-            - 30 normalized sequence-order correlation factors (theta values)
-
-        Raises
```
fkiraly, line 196: I think we should keep the "raises" block, no?

fkiraly (changes requested):
Looks good!

Comments are all about docstrings (removal, missing, or now incorrect content)

fkiraly (approved):
LGTM

PR #130, opened 2025-09-01
Add lightning wrapper for AptaTrans

### LINKED ISSUE #119: [ENH] `lightning` based AptaTrans model
In #112, we decided that `lightning` based models should be preferred.

@NennoMP also suggested to have `torch.nn` plus `lightning` models for each individual neural network.

Since AptaTrans currently is pure `torch.nn`, we should try this pattern out with AptaTrans and see whether we can derive an architectural principle, or whether we ought to adopt this as policy.

NennoMP:
This PR resolves #119.

Adds _lightning.LightningModule_ wrapper class to that `AptaTrans` may be trained using `lightning.Trainer` class.

Note that I have temporarily added an exclude `N812` in `pyproject.toml` because it considers bad code quality imports such as `import torch.nn.functional as F` or `import lightning as L`, as they're in upper-case. However, this is the style everyone uses for importing such modules (also the style used in their docs).

@fkiraly Making it compatible with `lightning` was actually cleaner and more straightforward then I expected. We only need one additional class `(AptaTransLightning)`. Opinions about the current design?


NennoMP:
Note that I have temporarily added an exclude `N812` in `pyproject.toml` because it considers bad code quality imports such as `import torch.nn.functional as F` or `import lightning as L`, as they're in upper-case. However, this is the style everyone uses for importing such modules (also the style used in their docs).


@fkiraly Making it compatible with `lightning` was actually cleaner and more straightforward then I expected. We only need one additional class `(AptaTransLightning)`. Opinions about the current design?

I think we will need another wrapper class for defining the training logic for the encoders alone, as it differs from the training logic for the neural network.

satvshr:
Just to add my 2 cents since I was advocating for lightning in the first place, this seems wayyyy better than the `AptaTransSolver` and `BaseSolver` needed for training. The `BaseSolver` (even though impressive) seems to be made completely redundant by `lightning`, which is enough reason to use `lightning` (peak propaganda).

satvshr:
> could you add links to the solvers?

I thought I had, updated the comment with links.

NennoMP:
> Just to add my 2 cents since I was advocating for lightning in the first place, this seems wayyyy better than the [`AptaTransSolver`](https://github.com/gc-os-ai/pyaptamer/pull/92/files#diff-b2c04b9628ac0434e2c3f0a417439b9136956e5b885d8a75b07c318942451f96) and [`BaseSolver`](https://github.com/gc-os-ai/pyaptamer/pull/92/files#diff-7bb9fe348adcff2df5503f8af8bc7eca603bcfa3086bd23ea6a416a1e59152a3) needed for training. The `BaseSolver` (even though impressive) seems to be made completely redundant by `lightning`, which is enough reason to use `lightning` (peak propaganda).

I totally agree. After giving a try I have to admit I changed my opinion about `lightning`. I think we can remove the solvers I implemented in #92.

PR #148, opened 2025-09-15
[ENH] pdb to String loader using SEQREQ records

### LINKED ISSUE #147: [ENH] `pdb` to `String` loader using `SEQREQ` records
Currently the only option to extract amino acids from a pdb file is to use 2 functions, `load_from_rcsb` and `struct_to_aaseq` like this:
```
from pyaptamer.datasets import load_from_rcsb
from pyaptamer.utils import struct_to_aaseq

print(struct_to_aaseq(load_from_rcsb("8UF8")))
```
So, using `load_from_rcsb` we convert from pdb to Structure and using `struct_to_aaseq` we convert from Structure to String, the problem with this approach is that biopython uses `ATOM` records from the pdb file as it contains 3d structure data about the pdb file. Hence, some parts of the sequence don't get captured because the 3 structure of that part is unknown. `SEQREQ` on the other hand contains the "simplified" version of the protein and outputs the "correct" sequence when extracted (as described in #138). Hence we should add a pdb to String converter as it would skip the conversion to Structure step and produce the complete sequence.

@KubiczekD @JaBirke please correct me if I am wrong.

satvshr:
Closes #147

This PR:
1. Adds a pdb to amino acid sequence transformation, without being lossy, following the convention mentioned in the 3 pointer list of [this](https://github.com/gc-os-ai/pyaptamer/issues/160#issuecomment-3362475358) comment.
2. Ensures `struct_to_aaseq` has an option to return a df too.

### DIFF
--- a/examples/aptanet_tutorial.ipynb
+++ b/examples/aptanet_tutorial.ipynb
@@ -54,7 +54,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 17,
+   "execution_count": null,
    "id": "a2f6701d",
    "metadata": {},
    "outputs": [],
@@ -68,7 +68,7 @@
     "]\n",
     "\n",
     "gnh = load_1gnh_structure()\n",
-    "protein_sequence = struct_to_aaseq(gnh, return_type=\"list\")\n",
+    "protein_sequence = struct_to_aaseq(gnh)\n",
     "\n",
     "# Build all combinations (aptamer, protein)\n",
     "X = [(a, p) for a in aptamer_sequence for p in protein_sequence]\n",
--- a/pyaptamer/utils/_pdb_to_aaseq.py
+++ b/pyaptamer/utils/_pdb_to_aaseq.py
@@ -56,7 +56,7 @@ def pdb_to_aaseq(pdb_file_path, return_type="list"):
     else:
         # Fall back to structure parsing
         structure = pdb_to_struct(pdb_path)
-        df = struct_to_aaseq(structure)
+        df = struct_to_aaseq(structure, return_type="pd.df")
 
     if df.empty:
         raise ValueError(f"No sequences could be extracted from PDB file: {pdb_path}")
--- a/pyaptamer/utils/_struct_to_aaseq.py
+++ b/pyaptamer/utils/_struct_to_aaseq.py
@@ -5,7 +5,7 @@ import pandas as pd
 from Bio.PDB.Polypeptide import PPBuilder
 
 
-def struct_to_aaseq(structure, return_type="pd.df"):
+def struct_to_aaseq(structure, return_type="list"):
     """
     Extract amino-acid sequences from a Biopython Structure.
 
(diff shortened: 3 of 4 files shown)


fkiraly (changes requested):
Does this function, at least ideas-wise, not do the same as `pdb_to_struct` then `struct_to_aaseq`? Recall that we separated the two intentionally, to separate file manipulation from in-memory objects which are abstract representations.

Instead of adding this new function which would link file manipulation back into the in-memory workflow, why not add features to the existing `struct_to_aaseq`?

satvshr:
> Instead of adding this new function which would link file manipulation back into the in-memory workflow, why not add features to the existing `struct_to_aaseq`?

https://github.com/gc-os-ai/pyaptamer/issues/147 answers your question, essentially 3d data is being captured when biopython converts something to a `Structure` object using ATOM records, sometimes a part of the sequence cna go missing if the 3d orientation of it is not known. This issue is not faced with SEQRES records, hence the direct converter.

pyaptamer/utils/pdb_to_aaseq.py
```diff
@@ -0,0 +1,43 @@
...
+def pdb_to_aaseq(pdb_file_path, return_df=False):
+    """
+    Extract amino-acid sequences (SEQRES) from a PDB file.
+
+    Parameters
+    ----------
+    pdb_file_path : str or os.PathLike
+        Path to a PDB file.
+    return_df : bool, optional, default=False
+        If True, return a pandas.DataFrame with columns:
```
fkiraly, line 19: let's try to be rst compatible where we can:

* newlines before and after bullet point lists
* double backticks around code like `'chain'` (note that markdown uses single backtick, rst wants double backtick)

pyaptamer/utils/pdb_to_aaseq.py
```diff
@@ -0,0 +1,43 @@
...
+
+def pdb_to_aaseq(pdb_file_path, return_df=False):
+    """
+    Extract amino-acid sequences (SEQRES) from a PDB file.
+
+    Parameters
+    ----------
+    pdb_file_path : str or os.PathLike
+        Path to a PDB file.
+    return_df : bool, optional, default=False
```
fkiraly, line 18: let's call this `return_type` with two possible values `"pd.df"` and `"list"`. This makes it more upwards compatible.

fkiraly:
> > Instead of adding this new function which would link file manipulation back into the in-memory workflow, why not add features to the existing `struct_to_aaseq`?
> 
> #147 answers your question, essentially 3d data is being captured when biopython converts something to a `Structure` object using ATOM records, sometimes a part of the sequence cna go missing if the 3d orientation of it is not known. This issue is not faced with SEQRES records, hence the direct converter.

follow-up question: would it be possible to get the sequences from the `Structure` object instead of the `pdb`?

fkiraly:
Ok, that is concerning. Are you saying the `Structure` objec tis *not* an 1:1 in-memory representation of the full `pdb`?

satvshr:
> Are you saying the `Structure` objec tis _not_ an 1:1 in-memory representation of the full `pdb`?

Just to have it on record, yes

satvshr:
Are chain ids which come along the amino acid sequences in the pdb file important too @JaBirke @KubiczekD ?

satvshr:
@rpgv could you please give an exmaple of a small pdb file with ATOM records but no SEQRES records? Need it for testing.

PR #153, opened 2025-09-21
[ENH] Update AptaNet notebook to use AptaTrans data and a pdb for prediction

### LINKED ISSUE #150: Update AptaNet notebook to use AptaTrans data and a pdb for prediction
currently, the AptaNet notebook uses randomly generated strings, i.e., dummy data.

We should improve the notebook so it shows how to use real data for training and inference.

As discussed, we should change the existing sections in the AptaNet notebook, to use as data the AptaTrans dataset, which is str x str -> binding 0/1.

We should also add two new sections:

1. where we train on the entire AptaTrans dataset, and predict binding probability (`predict_proba`) between a new pdb file (protein) and a DNA sequence
2. where we use MCTS combined with a trained AptaNet to propose new aptamers for a new pdb file, i.e., a form of in-silico Selex

satvshr:
closes #150


pyaptamer/datasets/_loaders/_li2014.py
```diff
@@ -0,0 +1,55 @@
...
+__all__ = ["load_train_li2014", "load_test_li2014"]
+import os
+
+import pandas as pd
+
+
+def load_train_li2014():
+    """
+    Load the Li 2014 training dataset.
+
```
fkiraly: please add some description of the dataset here. What are columns, what do they mean, what can the values be, etc.

pyaptamer/utils/_aptanet_utils.py
```diff
@@ -59,20 +60,18 @@ def generate_kmer_vecs(aptamer_sequence, k=4):
 def pairs_to_features(X, k=4):
```
fkiraly: I don't think this needs changing, this should be a transformer anyway

fkiraly (approved):
Great!

PR #155, opened 2025-09-23
[ENH] hugging face loader

### LINKED ISSUE #154: [DISCUSSION] Hugging face loaders
Adding a loader for hugging face data is turning out to be a slightly difficult task than expected,
@NennoMP tried implementing it as a part of #92 over [here](https://github.com/gc-os-ai/pyaptamer/pull/92/files#diff-15a3c9c6fa77cf579451b11e4b75a4135a8f7d8e3d402a8859f97b31f7698776), but there are a few problems I find with this loader:
1. We do not need to add an option to add local files, this is supposed to be a loader to upload files from the hugging face hub to in-memory objects.
2. [This](https://github.com/gc-os-ai/pyaptamer/pull/92/files#r2369899998)

After looking into it a bit more, hugging face only allows certain file formats (none of those being the ones we want like pdb, fasta, mmCIF, etc) and my solution for that is rather simple, this is the proposed code:
```
def hf_to_ds(path, keep_in_memory=True, **kwargs):
    """
    Load any Hugging Face dataset or file into a Dataset or DatasetDict.
    Tries native loader first; falls back to 'text' if unsupported.
    """
    try:
        return load_dataset(path, keep_in_memory=keep_in_memory, **kwargs)
    except Exception as e:
        # Fallback to text loader
        return load_dataset(
            "text",
            data_files=path,
            keep_in_memory=keep_in_memory,
            **kwargs
        )
```

For all possible data formats we first try calling it the "normal" way and if it does not work out we force the output to be in the form of a text file.
Now for the code in the except block, the [loader](https://huggingface.co/docs/datasets/v4.1.1/en/package_reference/loading_methods#datasets.load_dataset) takes a `split` argument and given pdb/fasta files do not have that, the function returns:
```
DatasetDict({
    train: Dataset({
        features: ['text'],
        num_rows: 652992
    })
})
```
as the default argument for `split` is 'train'. By adding:
```
    # Unwrap single-split DatasetDict into Dataset
    if isinstance(ds, dict) and len(ds) == 1:
        ds = next(iter(ds.values()))
```
to the except block we can get a [dataset](https://huggingface.co/docs/datasets/v4.1.1/en/package_reference/main_classes#datasets.Dataset) object throughout the entire function.

Hence the final code looks like:
```
def hf_to_ds(path, keep_in_memory=True, **kwargs):
    """
    Load any Hugging Face dataset or file into a Dataset or DatasetDict.
    Tries native loader first; falls back to 'text' if unsupported.
    """
    try:
        ds = load_dataset(path, keep_in_memory=keep_in_memory, **kwargs)
    except Exception:
        ds = load_dataset(
            "text",
            data_files=path,
            keep_in_memory=keep_in_memory,
            **kwargs
        )

    # Unwrap single-split DatasetDict into Dataset
    if isinstance(ds, dict) and len(ds) == 1:
        ds = next(iter(ds.values()))

    return ds
```

The problem with this is that it converts all data into a Hugging face dataset format which will be hard for researchers to understand. This is what needs to be done for FASTA files as an example using the above function:
```
dataset = hf_to_ds("https://huggingface.co/datasets/gcos/HoloRBP4_round8_trimmed/resolve/main/HoloRBP4_round8_trimmed.fasta")
fasta_string = "\n".join(dataset[:]['text'])

fasta_io = StringIO(fasta_string)

records = []
for record in SeqIO.parse(fasta_io, "fasta"):
    records.append({"id": record.id, "sequence": str(record.seq)})

return records
```

satvshr:
Fixes #154

### DIFF
--- a/pyaptamer/utils/__init__.py
+++ b/pyaptamer/utils/__init__.py
@@ -8,12 +8,10 @@ __all__ = [
     "rna2vec",
     "pdb_to_struct",
     "struct_to_aaseq",
-    "hf_to_dataset",
     "pdb_to_seq_uniprot",
     "pdb_to_aaseq",
 ]
 
-from pyaptamer.datasets._loaders._hf_to_dataset_loader import hf_to_dataset
 from pyaptamer.utils._aa_str_to_letter import aa_str_to_letter
 from pyaptamer.utils._pdb_to_aaseq import pdb_to_aaseq
 from pyaptamer.utils._pdb_to_seq_uniprot import pdb_to_seq_uniprot


pyaptamer/utils/fasta_to_aaseq.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Parameters
+    ----------
+    fasta_path : str or os.PathLike
+        Input source for FASTA sequences. Can be:
+          - Local file path (absolute or relative) located in 'pyaptamer/datasets/data'.
+          - Hugging Face Hub reference.
+    return_df : bool, default=False
+        If True, return a pandas DataFrame with columns:
+          - `id` : str
```
fkiraly, line 25: I think the `id` should be rows, and in the `index`

pyaptamer/utils/hf_to_dataset.py
```diff
@@ -0,0 +1,48 @@
+__author__ = "satvshr"
+__all__ = ["hf_to_dataset"]
+
+from datasets import load_dataset
+
+
+def hf_to_dataset(path, keep_in_memory=True, **kwargs):
```
fkiraly, line 7: this function feels unnecessary, it simply aliases a simple call of `huggingface` `load_dataset`.

Why do we need this? I would simply remove it.

satvshr: > Maybe you could explain?

The code in the try block is how you would normally call any supported file format by hugging face, if you do the same call for file formats not supported by hf, youll get an error. Hence in the except block we parse anything not suported by telling hf to recognise it as a text file. Now code will have to be repeated across all non-supported file format transofrmations like above in the `fasta_to_aaseq` function, which is:
```
        content = "\n".join(dataset[:]["text"])
        fasta_handle = StringIO(content)
```
which can then be used by biopython to convert it to our desired format.

fkiraly: is there not a condition that we can replace this in, in an if/else?

satvshr: If there was a defined variable which hugging face used to list all their recognisable file formats I would say yes, but as of now I dont think it is possible.

satvshr:
The initial ImportError was a masked error caused by naming a file the same as the function, so Python thought I was importing the file instead of the function. After renaming, the real problem showed up, a "circular import error" from python. Pytest starts by importing utils; inside __init__.py it first pulls in fasta_to_aaseq, which in turn tries to import hf_to_dataset from utils; but hf_to_dataset hasn’t been imported yet (it’s the next line in __init__.py), so Python throws the importerror. The fix is to either reorder imports so hf_to_dataset is defined first, or better, import it directly (which is what i did) to break the cycle.

pyaptamer/utils/_fasta_to_aaseq.py
```diff
@@ -0,0 +1,62 @@
...
+def fasta_to_aaseq(fasta_path, return_df=False):
+    """
+    Extract sequences from a FASTA file, Hugging Face dataset, or URL.
+
+    Parameters
+    ----------
+    ``fasta_path`` : str or os.PathLike
+        Input source for FASTA sequences. Can be:
+
+          - Local file path (absolute or relative) located in
```
fkiraly, line 22: is this all relative to the repo?

Note that users of the package will not have the repo cloned. Please change this so it is relative to, say, root, and/or that absolute paths not subpaths of the repo can be passed.

pyaptamer/utils/_hf_to_dataset.py
```diff
@@ -0,0 +1,48 @@
+__author__ = "satvshr"
+__all__ = ["hf_to_dataset"]
+
+from datasets import load_dataset
+
+
+def hf_to_dataset(path, keep_in_memory=True, **kwargs):
```
fkiraly: I still doubt the reasoning behind having this function.

Can you please give two examples where we currently need this, one example for each branch of the code?

pyaptamer/utils/_fasta_to_aaseq.py
```diff
@@ -0,0 +1,62 @@
...
+import pandas as pd
+from Bio import SeqIO
+
+from pyaptamer.utils._hf_to_dataset import hf_to_dataset
+
+
+def fasta_to_aaseq(fasta_path, return_df=False):
+    """
+    Extract sequences from a FASTA file, Hugging Face dataset, or URL.
+
```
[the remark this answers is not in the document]
satvshr: > the exact format the URL is needed (is this actually an option or just hallucinated?)

It is not hallucinated; load_dataset can handle URLs.

pyaptamer/utils/_fasta_to_aaseq.py
satvshr: Very rookie mistake on my end 😓

pyaptamer/utils/_hf_to_dataset.py
fkiraly: can you please give an example of a scenario where this would be useful?

satvshr: > `try` block: We want to load a file from hugging face which is in one of the known file formats supported by hugging face, so something like the [imbd](https://huggingface.co/datasets/stanfordnlp/imdb) dataset, which is a csv file (known file format).
> 
> `except` block: We want to load a file from hugging face which is not one of the known file formats (say a fasta/pdb/mmcif file), like the fasta file which Jakob provided. If it was used in the above `try1 block, it would throw an error. This block will also work if any local files are supplied to the loader and would return a Dataset object.

The `try` case would need a call like `load_dataset("imbd")` whereas in the `except` case it would need a call like `load_dataset("text", data_files="https://huggingface.co/datasets/gcos/HoloRBP4_round8_trimmed")`

pyaptamer/trafos/convert/_any_to_aaseq.py
```diff
@@ -0,0 +1,79 @@
...
+            DataFrame with a single column ``"sequence"``, where each row
+            corresponds to one parsed sequence as a string.
+
+        Raises
+        ------
+        ValueError
+            If the specified format is not supported or parsing fails.
+        """
+        file_path = self._check_X(X)
+        records = [
```
fkiraly: this is very nice - but I think it should be integrated into `MoleculeLoader` and not be a transformer.

* `_determine_type` based on file ending, for the formats for which file ending is known.
* if the type is determined to be one supported by `SeqIO`, then call a catch-all function `_load_seqio(self, path, format)`, were `format` is passed to `SeqIO.parse` (second argument)

fkiraly (changes requested):
I left some comments above:

* on "to asseq" feature, this should move to `MoleculeLoader`
* the hf loader is ok
* I would also add another dispatch path to `MoleculeLoader`, where a path `hf://gcos/Holo` etc is admissible. Separate PR perhaps.

fkiraly:
> @fkiraly should we assume that each pdb file is only going to return one sequence? That is what you are doing in `_load_pdb_seq` of `MolecularLoader` by only returning one string right?

is that a bad assumption? Could be, please explain.

PR #158, opened 2025-09-25
[ENH] aptacom database loader 

rpgv:
Added: AptaCom loader - Loads AptaCom dataset from hugging face in one of two formats: 
pandas dataframe, hugging face dataset;

### DIFF
--- a/pyaptamer/datasets/_loaders/_aptacom_loader.py
+++ b/pyaptamer/datasets/_loaders/_aptacom_loader.py
@@ -127,7 +127,7 @@ def load_aptacom_xy(return_X_y=False):
     --------
     Either a pandas dataframe with three columns
     or two pandas dataframe objects with two and one
-    columns respectively. 
+    columns respectively.
     """
     aptacom = load_hf_dataset("AptaCom", store=False)
     dataset = prepare_xy(aptacom)


pyaptamer/datasets/_loaders/_aptacom_loader.py
```diff
@@ -0,0 +1,77 @@
...
+from datasets import load_dataset
+
+
+def load_aptacom(as_df=False, filter_entries=None):
+    """
+    Loads complete aptacom dataset from hugging face datasets.
+
+    Parameters
+    ----------
+    as_df: (bool) Requires pandas compatible format; converts
```
fkiraly, line 14: defaults should be stated in the docstring, and please adopt a consistent formatting. Suggestion:

`paramname : type/description, default=default_value`

pyaptamer/datasets/_loaders/_aptacom_loader.py
```diff
@@ -0,0 +1,77 @@
...
+    dataset into pandas dataframe
+
+    filter(str): Allows filtering dataset by:
```
fkiraly, line 17: parameter is called differently in function signature

satvshr:
> > so should this be paused till then so that it can be updated with the Hugging Face loader we will be using?
> 
> I do not think we should block this. I would instead refactor, keeping the public interfaces the same.

I left some comments, could you have a look and tell me why they are not something that should be changed so that I do not raise such points again?

pyaptamer/datasets/_loaders/_aptacom_loader.py
```diff
@@ -0,0 +1,138 @@
...
+        ],
+    ),
+}
+
+
+def filter_columns(ds, columns=None):
+    """ " Selects columns to keep on dataset
+    Parameters:
+    -----------
+        ds: pd dataframe, required
```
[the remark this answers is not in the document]
satvshr: Check #192 :D

PR #165, opened 2025-09-30
[ENH] making `AptaTrans` sklearn-like

### LINKED ISSUE #190: [ENH] Changes to `AptaTransPipeline` to make it sklearn-like
The current design of `AptaTransPipeline` appears inconsistent with the expected scikit-learn–style API since it lacks `fit` and `predict` methods. Without these common interfaces, it’s unclear how benchmarking is intended to operate, especially given that `AptaNetPipeline` does follow this pattern: there’s no shared method that can be called on both classes for fitting and generating predictions. Hence, we should add `fit` and `predict` methods to `AptaTransPipeline`.

Additionally:
1. The description of `prot_words` should be standardized and rewritten in a clear, generic way across all instances where it appears, as its current form is overly confusing.
2. The `predict` function should explicitly document the datatype returned within the Tensor; at present, users have to inspect the `evaluate` function to infer what is being returned.

satvshr:
closes #190

### DIFF
--- a/pyaptamer/aptatrans/_pipeline.py
+++ b/pyaptamer/aptatrans/_pipeline.py
@@ -136,13 +136,12 @@ class AptaTransPipeline(BaseObject):
     def _init_aptamer_experiment(self, target: str) -> AptamerEvalAptaTrans:
         """Initialize the aptamer recommendation experiment."""
         self._init_vocabularies()
-        experiment = AptamerEvalAptaTrans(
+        return AptamerEvalAptaTrans(
             target=target,
             model=self.model,
             device=self.device,
             prot_words=self.prot_words_,
         )
-        return experiment
 
     def _preprocess_data(self, dataset) -> DataLoader:
         """Convert numpy arrays or pandas DataFrames into a PyTorch DataLoader."""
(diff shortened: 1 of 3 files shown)


satvshr:
@NennoMP for whenever you have time, given I think I have done most of what I could do:
1. Check out `predict`as it doesnt use `Trainer.test` at the moment it uses MCTS' `recommend`.
2. If you run the current benchmarking notebook you will get this error: `RuntimeError: The size of tensor a (0) must match the size of tensor b (128) at non-singleton dimension 1` which is something to look into.
3. Update tests to present pipeline implementation if you find current work satsfying enough.

NennoMP:
> @NennoMP for whenever you have time, given I think I have done most of what I could do:
> 
> 1. Check out `predict`as it doesnt use `Trainer.test` at the moment it uses MCTS' `recommend`.
> 2. If you run the current benchmarking notebook you will get this error: `RuntimeError: The size of tensor a (0) must match the size of tensor b (128) at non-singleton dimension 1` which is something to look into.
> 3. Update tests to present pipeline implementation if you find current work satsfying enough.

Thank you. I'll try to look into it this week.

Where is bug (2.) occurring specifically, an you provide the traceback? The AptaTrans notebook is working on my side, so the problem could be in the benchmark class and/or benchmark notebook.

PR #178, opened 2025-10-17
[BUG] AptaTrans bug fixes: shapes mismatch in pretraining, add masking in encoders

### LINKED ISSUE #177: [BUG] AptaTrans: shapes mismatch during pretraning and attention computed on padded zeros
There are a few problems in AptaTrans model/training: (i) shape mismatch between predictions and ground-truth labels when pretraining the encoders, and (ii) attention is currently being computed also on padded-zeros. Specifically, the former prevents correct training of the encoders, while the latter is useless from a computational perspective and may also degrade performance.

This is a blocker for #146.

NennoMP:
This PR resolves #177.

It addressed (i) a bug related to a shape mismatches between predictions and ground-truth labels when pretraining the encoders. Also, (ii) added an attention mask to avoid attention looking at padded zeros (as in the original implementation). This avoid useless computations and potential degradation of performance.

Consequently, a few other methods in the utils and/or experiments had to be updated. Similarly, docstrings and tests have been updated.

In short, the changes are the following:
* In `pyaptamer.aptatrans._model_lightning`, specifically its `training_step(...)` method. This solves one of the shape mismatche with a simple `transpose`;
* In `pyaptamer.utils._rna` and `pyaptamer.utils._aptatrans_utils`. These changes are related to the update of the `generate_nplets(...)` method. The old method was mapping n-plets to incorrect unique token integers.
* In `pyaptamer.aptatrans._model`, specifically its `make_encoder(...)` method. This change adds the attention mask and passes it to the encoder;

After adding the attention mask the docstrings of `AptamerEvalAptaTrans` and `MCTS` classes were failing. I realized that `prot_words` was supposed to map n-plets to integer IDs rather than float: while `prot_words` indeed starts as a mapping between n-plets and float frequencies, the embeddings in the encoders expect them mapped to integers. I updated the docstrings accordingly. Changes in `pyaptamer.experiments._aptamer` and `pyaptamer.MCTS._algorithm` are related to this.

### DIFF
--- a/pyaptamer/aptatrans/_model.py
+++ b/pyaptamer/aptatrans/_model.py
@@ -185,7 +185,7 @@ class AptaTrans(nn.Module):
 
         Returns
         -------
-        nn.Sequential
+        nn.ModuleList
             A sequential container with the encoder's architectural components.
         TokenPredictor
             A token predictor layer for masked token and secondary structure prediction.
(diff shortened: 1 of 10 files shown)


fkiraly (approved):
Looks good to me.

I find it a little concerning though that we did not detect this previously. From our discussion, that is due to AptaTrans not being tested with a full workflow, which in turn is due to significant inference time.

I would recommend: open an issue about test coverage of AptaTrans and its significant inference time.

PR #192, opened 2025-11-01
[DOC] improve aptacom loader docstrings

fkiraly:
improves aptacom loader docstrings - actions the comments by @satvshr that I overlooked in https://github.com/gc-os-ai/pyaptamer/pull/158.

Sorry for missing these and merging.

### DIFF
--- a/examples/aptatrans_tutorial.ipynb
+++ b/examples/aptatrans_tutorial.ipynb
@@ -49,7 +49,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "BATCH_SIZE = 16\n",
     "TEST_SIZE = 0.05  # size of the test set for pretraining\n",
     "RAMDOM_STATE = 42  # for reproducibility\n",
     "\n",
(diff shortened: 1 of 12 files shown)


satvshr:
Missing it was fine and totally understandable, thanks for making an issue and trying to fix the change requests! Could you tell me why these changes should not be made for future purposes:
1. [Link](https://github.com/gc-os-ai/pyaptamer/pull/158#discussion_r2467036357)
2. [Link](https://github.com/gc-os-ai/pyaptamer/pull/158#discussion_r2467041307)

pyaptamer/.DS_Store
satvshr: what is this file?

fkiraly:
> Missing it was fine and totally understandable, thanks for making an issue and trying to fix the change requests! Could you tell me why these changes should not be made for future purposes:

> should we not name all functions with x and y in their name similarly? So this would be prepare_x_y instead.

I think it does not matter since it is not intended to be a public function (and should perhaps start with a `_` hence)

> Should this function not be private? Also do you mind changing `ds` to `df` (for dataframe) given `ds` is something I will be using for the `Dataset` object from HuggingFace?

Actually, yes.

PR #198, opened 2025-11-21
[ENH] Add 5NU7 and 1BRQ

### LINKED ISSUE #197: [ENH] Add 5NU7 and 1BRQ
Adding 5NU7 and 1BRQ pdb file and loaders, given that is what is mostly used by the corporation partners of ecoSPECS. I am going to directly add them to the library as they are about 200kb in size.

satvshr:
Fixes #197.

from the issue description: Adding 5NU7 and 1BRQ pdb file and loaders, given that is what is mostly used by the corporation partners of ecoSPECS. I am going to directly add them to the library as they are about 200kb in size.


fkiraly (changes requested):
I think you can already see by the fact that so many locations are affected that something isn't quite right.

* can you add loaders `load_5nu7` and `load_1brq` to be used in example code?
* Why are you just deleting the 1GNH and PFOA loaders? I see no reason for that, add them back please.
* the function `structure_loader` feels unnecessary. It just aliases `PDBParser.get_structure`. Youcan also see that it was a bad idea since you need to use unnevessary `lambda` expressions to make current code work. I suggest: remove.
* the function `mol_loader` feels unnecssary. It just aliases `MoleculeLoader.__init__`.  It would be useful of all `pdb` strings could be used, but currently only two can? I suggest: remove.

fkiraly (changes requested):
please add it back because we have been using it in testing. Thanks.

(this is now the third round where I ask you to add it back and not delete it, and you "react" by still deleting it)

fkiraly (changes requested):
Conflicting with `main` after recent merges - kindly resolve

satvshr:
Logs are no longer visible (?) for the failed run given you reran the tests, learnt a lesson to always dummy commit instead of rerunning tests to preserve the test logs 😅

PR #209, opened 2025-12-01
[ENH] Remove `Structure` support for packaged pdbs

### LINKED ISSUE #208: [ENH] Remove `Structure` support for packaged pdbs
Now that we are using `MoleculeLoader` as our default in-memory format, we can remove the `Structure` loader for all packaged pdbs. We are not completely removing `Structure` support, though, as we still have functions in `utils` to convert from pdb to struct and struct to aaseq.

satvshr:
stacks on #198 
fixes #208 
Now that we are using `MoleculeLoader` as our default in-memory format, we can remove the `Structure` loader for all packaged pdbs. We are not completely removing `Structure` support, though, as we still have functions in `utils` to convert from pdb to struct and struct to aaseq.


pyaptamer/datasets/_loaders/_1gnh.py
```diff
@@ -19,7 +19,8 @@ def load_1gnh():
     return MoleculeLoader(pdb_path)
 
 
-def load_1gnh_structure(pdb_path=None):
+# This function is provided only to test struct_to_aaseq.
+def _load_1gnh_structure(pdb_path=None):
```
satvshr: Function kept as it is used for testing `struct_to_aaseq`

siddharth7113:
Also this stack #198 which @fkiraly needs to approve, then we can merge this one,

PR #214, opened 2025-12-08
Feature/190 make aptatrans sklearn compatible

### LINKED ISSUE #190: [ENH] Changes to `AptaTransPipeline` to make it sklearn-like

NennoMP:
This PR resolves #190.

Added `fit(...)` and `predict(...)` methods to `AptaTransPipeline` to make it sklearn compatible. Here is a summary of the changes to facilitate review:

* Changes to the **pipeline** (`_pipeline.AptaTransPipeline`):
  * Added a `fit(...)` method with sklearn-like interface to enable training;
  * Added a `predict(...)` method with sklearn-like interface to enable prediction.
  
* Changes to **lightning training code** (`_model_lightning.AptaTransLightning`, `_model_lightning.AptaTransEncoderLightning`):
  * Added a `predict_step(...)` method to enable inference when ground-truth labels are not available (since sklearn predict interface does not expect them);
  * Modified `_step(...)` method accordingly to handle both cases when ground-truth labels are or are not provided.
  
* Changes to **dataloaders** (`_api.APIDataset`):
  * Similarly to above, changes were needed to handle the scenario where ground-truth labels are not provided/available.
  
Finally, tests were updated accordingly.


pyaptamer/aptatrans/_model_lightning.py
```diff
@@ -76,31 +75,44 @@ def __init__(
         self.weight_decay = weight_decay
         self.betas = betas
 
-    def _step(self, batch: tuple[Tensor, Tensor, Tensor], batch_idx: int) -> Tensor:
-        """Defines a single (mini-batch) step in the training/test loop.
+    def _step(
+        self,
+        batch: tuple[Tensor, Tensor] | tuple[Tensor, Tensor, Tensor],
+        batch_idx: int,
```
siddharth7113, line 81: Why are we passing `batch_idx` , I couldn't find the use in this particular function?

pyaptamer/aptatrans/_model_lightning.py
```diff
@@ -143,13 +155,35 @@ def test_step(self, batch: tuple[Tensor, Tensor, Tensor], batch_idx: int) -> Ten
...
         """
-        loss, accuracy = self._step(batch, batch_idx)
+        _, loss, accuracy = self._step(batch, batch_idx)
 
         self.log("test_loss", loss, on_epoch=True, on_step=False, prog_bar=True)
         self.log("test_accuracy", accuracy, on_epoch=True, on_step=False, prog_bar=True)
 
         return loss
 
+    def predict_step(
```
siddharth7113, line 165: Would this function act as a replacement of test_step,

instead of using `test_step` to calculate loss, you call `predict_step` use it's output and ground_truth and then calculate loss , optimizier, backward pass and so on?

If so, can you open an issue about the same?

pyaptamer/aptatrans/_model_lightning.py
```diff
@@ -256,45 +291,58 @@ def __init__(
         self.weight_mlm = weight_mlm
         self.weight_ssp = weight_ssp
 
-    def _step(self, batch: tuple[Tensor, Tensor, Tensor], batch_idx: int) -> Tensor:
-        """Defines a single (mini-batch) step in the training/test loop.
+    def _step(
+        self,
+        batch: tuple[Tensor, Tensor] | tuple[Tensor, Tensor, Tensor, Tensor],
+        batch_idx: int,
```
siddharth7113, line 297: batch_idx is not being used here in the function

siddharth7113 (commented):
Overall looks good for now, but at later step, instead of having a test step we can use predict and use a seaprate loss calculation and optimiser loop like pytorch? (suggestion)

apart from couple of comments about `batch_idx`, this could be merged for now.

siddharth7113:
@NennoMP  since this PR has been pending from long time and the structure has changed since then, could we close this and add instead detailed description in issue , open for new contributors.

NennoMP:
@siddharth7113 I think so. A lot of stuff changed so it may be better to start fresh I guess? I would still reference this since I think the changes to the pipeline are still relevant/good.

siddharth7113:
> @siddharth7113 I think so. A lot of stuff changed so it may be better to start fresh I guess? I would still reference this since I think the changes to the pipeline are still relevant/good.

Agreed, can you just make a new issue , mention this PR there, along with all the references and details, and maybe add labels ,good-first-issue` and probably `medium` or `hard` , otherwise there would be a in-flux of PR on this.

PR #217, opened 2025-12-11
[BUG] `AptaNetPipeline` failing tests

### LINKED ISSUE #216: [BUG] `AptaNetPipeline` failing tests
With scikit-learn releasing 1.8 (an upgrade from 1.7.1), `AptaNetPipeline` is failing tests with the error being:
```
AttributeError: The following error was raised: 'AptaNetPipeline' object has no attribute '__sklearn_tags__'. It seems that there are no classes that implement `__sklearn_tags__` in the MRO and/or all classes in the MRO call `super().__sklearn_tags__()`. Make sure to inherit from `BaseEstimator` which implements `__sklearn_tags__` (or alternatively define `__sklearn_tags__` but we don't recommend this approach). Note that `BaseEstimator` needs to be on the right side of other Mixins in the inheritance order.
```

satvshr:
Closes #216 

This PR restricts the use of sklearn version 1.8 given it causes failures in `AptaNetPipeline `and imblearn (used in the `AptaNetPipeline` notebook). It also adds `BaseEstimator` as a dependency to `AptaNetPipeline` to suppress warnings.

---

An earlier bug report (NOT RELEVANT AT THE MOMENT) but it is interesting that skorch version 1.2 and sklearn version 1.8 stop causing errors in AptaNetPipeline (though imblearn still fails). It is also worth noting that the new skorch version was not causing problems given CI tests were passing with skorch version 1.3 (latest) and sklearn version 1.7.x:
To be compatible with sklearn 1.8 all that was needed was adding `BaseEstimator` to `AptaNetPipeline` (tests started passing locally when I did this, the CI tests had lesser failures as compared to the ones observed on main: [main failures](https://github.com/gc-os-ai/pyaptamer/actions/runs/20133669478/job/57781221445) vs [after adding `BaseEstimator`](https://github.com/gc-os-ai/pyaptamer/actions/runs/20147077744/job/57830282558?pr=217) you will observe 38 vs 36 failed tests), but CI tests continue failing even though at that point locally all tests were passing. After I updated my local versions of skorch, numpy and scikit-base to match the ones being used by the CI, I started getting the same errors as CI is getting locally. To isolate the problem I then downgraded my version of skorch and I discovered that the newest version of skorch was causing these problems.

TLDR: Bug was not only because of sklearn new version but mainly skorch. I suggest having skorch version 1.2 to be the one we use instead of the latest 1.3, but that is just what I suggest I do not know if thats best practice or not. I will continue looking into the error, though I cannot find a solution at the moment, and given it involves using a skorch model, inside a pipeline, inside a sklearn-compatible estimator; I do not think it is a simple fix.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,8 +21,8 @@ dependencies = [
     "pandas>=2.0.0",
     "scikit-base",
     "torch>=2.5.1",
-    "scikit-learn>=1.3.0",
-    "skorch<=1.2.0",
+    "scikit-learn>=1.3.0,<1.8.0",
+    "skorch",
     "imblearn",
     "datasets",
 ]


satvshr:
This PR restricts the use of sklearn version 1.8 given it causes failures in `AptaNetPipeline `and imblearn (used in the `AptaNetPipeline` notebook). It also adds `BaseEstimator` as a dependency to `AptaNetPipeline` to suppress warnings.

PR #220, opened 2025-12-17
Add ValueError when AptaTransPipeline receives a too small depth value

### LINKED ISSUE #196: [BUG] AptaTrans aptamer recommendation does not work for depth smaller than 3
Currently, running aptamer recommendation with AptaTrans (i.e., combining it with MCTS) does not work when the depth of the tree search is set to values `1` or `2`. 

The reason is that AptaTrans pre-processing involves triplet encoding (i.e., uses 3-mers). Specifically, this happens when we call `_rna.rna2vec(...)` in `_experiment_aptatrans.AptamerEvalApTrans`. Therefore, when a sequence of length 1 or 2 is provided the method returns an empty tensor which causes the forward pass of the neural network to fail.

If you run the following code with `depth=1` or `depth=2`:

```python
pipeline = AptaTransPipeline(
    device=device,
    model=model,
    prot_words=prot_words,
    depth=1,
    n_iterations=1, 
)
candidates = pipeline.recommend(
    target=target_protein,
    n_candidates=1,
    verbose=True,
)
```

You get:

```
File \pyaptamer\pyaptamer\aptatrans\layers\_encoder.py:85, in PositionalEncoding.forward(self, x)
     68 """Forward pass.
     69 
     70 Parameters
   (...)     79     positional encodings applied.
     80 """
     81 assert x.shape[1] <= self.max_len, (
     82     f"Input sequence length {x.shape[1]} exceeds maximum length {self.max_len}."
     83 )
---> 85 out = x + self.pe[:, : x.shape[1], :]
     86 if self.dropout:
     87     out = self.dropout(out)

RuntimeError: The size of tensor a (0) must match the size of tensor b (128) at non-singleton dimension 1
```

NennoMP:
This PR resolves #196.

Added a `raise ValueError(...)` when the `AptaTransPipeline` is initialized with `depth` smaller than 3. Therefore, changes are limited to `aptatrans.pipeline.AptaTransPipeline` and its corresponding tests, where I added a new test to check whether the exception is correctly raised.

See the issue updated description for more info about why the bug occurred. I believe the best option is to raise an exception because the triplet (3-mers) encoding used by AptaTrans will simply produce empty vectors when provides sequences with length less than 3. This encoding is used in the original representation.

### DIFF
--- a/examples/aptatrans_tutorial.ipynb
+++ b/examples/aptatrans_tutorial.ipynb
@@ -548,7 +548,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.9"
+   "version": "3.11.13"
   }
  },
  "nbformat": 4,
--- a/pyaptamer/aptatrans/_pipeline.py
+++ b/pyaptamer/aptatrans/_pipeline.py
@@ -43,7 +43,10 @@ class AptaTransPipeline:
         subsequences and their frequency should come from the same dataset used for
         pretraining the protein encoder.
     depth : int, optional, default=20
-        The depth of the tree in the Monte Carlo Tree Search (MCTS) algorithm.
+        The depth of the tree in the Monte Carlo Tree Search (MCTS) algorithm. Also
+        defines the length of the generated aptamer candidates. Must be equal or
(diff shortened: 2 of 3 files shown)


siddharth7113:
lgtm

PR #231, opened 2026-02-16
[ENH] `MoleculeLoader` support for `PSeAAC`

### LINKED ISSUE #46: PSeAAC on in-memory `MoleculeLoader`
Once we introduce `MoleculeLoader` for representing proteins or amino-acid sequences, we should ensure that this is also a valid input for `PSeAAC`.

Steps to add this in a programmatic way:

* the boilerplate coerces `MoleculeLoader` to `str` by extracting the amino-acid sequence
* add tests for `MoleculeLoader` input

pandey-ps:
resolves #46 

Added `MoleculeLoader` as input type for `PseAAC` and `AptaNetPseAAC`, for the string input it is parsed as before (returns 1D array), and for `MoleculeLoader` (returns list of 1D arrays)

-tests to be added


fkiraly (changes requested):
Great! Thanks for contributing!

To align this with the current design: you should inherit `AptaNetPSeAAC` from `BaseTransform`.

Feel free to suggest minimal modifications to `BaseTransform` if it does not fit.

I would also move the PSeAAC algorithms to the `trafos` folder.

prashantpandeygit:
for the `BaseTransform`, it plays with dataframe as input and output, so the `_aptanet_utils.py` needs to wrap each string into a dataframe, or should we still be accepting strings only? or better we could accept strings also and auto wrap them into a dataframe?

also tests for `AptaNetPSeAAC` needs to be changes to address dataframe .

siddharth7113:
Hi @prashantpandeygit CI is failing currently, I would recommend to run pre-commit on the modified files.

satvshr:
@prashantpandeygit Kindly install `pre-commit` in your env before pushing changes.

prashantpandeygit:
I did run it before, but forgot when moved algorithms to trafos folder. my bad, will do it.

fkiraly (changes requested):
Same as above:

* the transformers should inherit from `BaseTransform`
* `transform` should not get overridden

prashantpandeygit:
`transform` does inherit from `BaseTransform`, also neither `transform` nor `_transform` is overriden , it only overrides `_transform_element` (private).

cc @satvshr @siddharth7113

fkiraly:
No, it does not. Not all transformers (the classes) you worked on inherit from `BaseTransform` now.
Look at `PSeAAC`.

prashantpandeygit:
I did not touch `PSeAAC` as i thought modifications were only needed for `AptaNetPSeAAC` as said [here](https://github.com/gc-os-ai/pyaptamer/pull/231#pullrequestreview-3809486073), will do the changes for PSeAAC now.

pyaptamer/trafos/pseaac/_pseaac_general.py
```diff
@@ -219,12 +229,6 @@ def transform(self, protein_sequence):
         protein_sequence : str
             The input protein sequence consisting of valid amino acid characters
             (A, C, D, E, F, G, H, I, K, L, M, N, P, Q, R, S, T, V, W, Y).
-        lambda_val : int, default=30
```
pandey-ps, line 222: these parameters (`lambda_val` and `weight` were not a part of the method, so removed the docstring. (they are used with `self`).

prashantpandeygit:
now both the algorithms inherit from `BaseTransform`

pyaptamer/trafos/base/_base.py
```diff
@@ -96,6 +96,13 @@ def _transform(self, X):
             Transformed data.
         """
         if self.get_tag("property:elementwise", False):
+            if isinstance(X, str):
```
siddharth7113, line 99: Why is the check happening here? Should it not be placed in `_check_X_y`?

_transform should not deal with this.

pyaptamer/trafos/pseaac/_pseaac_general.py
```diff
@@ -114,6 +115,13 @@ class PSeAAC:
     [0.006 0.006 0.006 0.006 0.006 0.006 0.018 0.018 0.018 0.018]
```
siddharth7113, line 115: Shouldn't all trafos return type should be same i.e. `pd.dataframe`

pyaptamer/trafos/base/_base.py
```diff
@@ -149,6 +156,10 @@ def _check_X_y(self, X, y):  # noqa: N802
         """
         if isinstance(X, MoleculeLoader):
             X = X.to_df_seq()
+
```
siddharth7113, line 156: instead of doing this, why can't we keep the abstract data type as `df` :
```python
def _check_X_y(self, X, y):
    if isinstance(X, MoleculeLoader):
        X = X.to_df_seq()
    elif isinstance(X, str):
        X = pd.DataFrame({"sequence": [X]})         # single-row df
    elif isinstance(X, list):
        X = pd.DataFrame({"sequence": X})           # n-row df

    # Now the original 3 checks run normally
    if not isinstance(X, pd.DataFrame):
        raise TypeError(...)
    if not self.get_tag("capability:multivariate") and X.shape[1] > 1:
        raise ValueError(...)
    return X, y

````

siddharth7113 (changes requested):
Hi, 

I have left the comments , and feedback on the current design, it needs non-trivial changes.

pyaptamer/trafos/base/_base.py
pandey-ps: oh yes input type shouldn't be checked here

pyaptamer/trafos/pseaac/_pseaac_general.py
pandey-ps: I agree, i did ask for it before [here](https://github.com/gc-os-ai/pyaptamer/pull/231#issuecomment-3912906351), so should I make this change in `BaseTransform` ?

pyaptamer/trafos/base/_base.py
pandey-ps: we would need to change tests for this too.

pyaptamer/trafos/pseaac/_pseaac_general.py
siddharth7113: Yes, also I have opened #441 to refactor , I have kept this PR in mind while refactoring, you can look up there, how going forward this transformation would fit.

prashantpandeygit:
@siddharth7113 I had a look at the PR, the changes here to make would only be to the algorithms adding a dataframe wrapper, and make changes in tests. as `BaseTransform` is being taken care at PR #441 . should I wait for its merger?

siddharth7113:
No, I have closed that PR at the moment, we would be modifying the `MoleculeLoader` , you can continue working on this one. We would be coming back to figure this out once we have finalised design for it.

pandey-ps:
> No, I have closed that PR at the moment, we would be modifying the `MoleculeLoader` , you can continue working on this one. We would be coming back to figure this out once we have finalised design for it.

Could you help how should I approach this now as algorithms now use `BaseTransform`, so what is here to do more in this PR? adding dataframe only as input and output?

PR #233, opened 2026-02-22
fix: remove redundant inits from _loaders/__init__.py  (resolves #225)

### LINKED ISSUE #225: [INVALID] remove all inits from `pyaptamer\datasets\_loaders\__init__.py`
`pyaptamer\datasets\_loaders\__init__.py` is basically a mirror of `pyaptamer\datasets\__init__.py`, and every time I add a new loader, I have to add it to both the above-mentioned inits. Given `pyaptamer\datasets\_loaders\__init__.py` import will never be used (given ` _loaders` is private) can we just not have any inits in `pyaptamer\datasets\_loaders\__init__.py`?

Dreamstick9:
this PR resolves #225 by removing all redundant `import` statements and the `__all__` list from `pyaptamer/datasets/_loaders/__init__.py`. 

As noted in the issue, `_loaders` is a private directory, and exposing its contents in its own `__init__.py` only created unnecessary double-work when adding new dataset loaders to the project.please note that in  this PR i did not delete the file. 

Closes #225

### DIFF
--- a/pyaptamer/datasets/_loaders/__init__.py
+++ b/pyaptamer/datasets/_loaders/__init__.py
@@ -1,1 +1,1 @@
-"""Loaders for different data structures."""
+"""Private loaders module. Imports are exposed via `pyaptamer/datasets/__init__.py`."""


Dreamstick9:
@satvshr  could you please review this PR?

PR #245, opened 2026-03-08
fix(utils): use os.path.basename for PDB structure ID

tkshsbcue:
## Summary

Replaces manual path splitting with `os.path.basename()` when deriving the structure identifier in `pdb_to_struct`.

## Changes

**`_pdb_to_struct.py`**

The previous logic used `pdb_file_path.split("/")[-1].split("\\")[-1]` to get the filename for the Biopython structure ID. That:

- Assumes only `/` and `\\` as path separators
- Fails on mixed or non-standard separators
- Ignores `os.path` conventions for cross-platform paths

Switching to `os.path.basename(pdb_file_path)` uses the standard library and behaves correctly on all supported platforms.

### DIFF
--- a/pyaptamer/utils/_pdb_to_struct.py
+++ b/pyaptamer/utils/_pdb_to_struct.py
@@ -1,6 +1,8 @@
 __author__ = "satvshr"
 __all__ = ["pdb_to_struct"]
 
+import os
+
 from Bio.PDB import PDBParser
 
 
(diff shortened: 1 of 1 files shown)


satvshr:
I don't see a problem with what's there currently, this doesn't seem necessary

siddharth7113:
@satvshr  I would say it would increase code readability? If the code is correct, we can merge it?

satvshr:
Tests seem to be passing so I don't think the code is wrong, can be merged if you feel so.

Issue #170, opened 2025-10-09
Sklearn: string transformations


NennoMP:
As highlighted in #106, we need to structure the string transformations currently present in the package under a consistent interface compatible with `sklearn`.

Below the tenative design suggested by @fkiraly in https://github.com/gc-os-ai/pyaptamer/issues/106#issuecomment-3256299076:

`BaseStrTransformer` 

* follows `sklearn` contract elements: dataclass-like `__init__`
* inherits from `skbase` `BaseObject`
* with methods as below

`fit(X: pd.DataFrame) -> self`

* `X` has `str`-valued columns, and can have other columns with float or int value
* by default empty

`transform(X: pd.DataFrame -> pd.DataFrame)`

* `X` has `str`-valued columns, and can have other columns with float or int value
* must be implemented

tags:

* `transform:output` can be `str` or `float`, depending on the output of `transform`
* `capability:multivariate`: whether input `X` in `transform` can have multiple columns

PR #246, opened 2026-03-08
Add torch-compatible transformations

### LINKED ISSUE #171: PyTorch: string transformations
As highlighted in #106, we need to structure the string transformations currently present in the package under a consistent interface compatible with `torch` dataloaders.

We should implement these transformations under `torch` core. There is not an established interface as transformations are generally contained elsewhere, e.g. `torchtext` or `torchvision` for text and images, respectively. Thus, rather than inheritance, the design is currently organized with duck typing - transforms should implement the `__call__` method in order to work with `nn.Sequential`. Note sure whether it is better to define an abstract class for the interface or stick with duck typing though.

Unfortunatelty `torchtext` stopped development as of Sep. 2023, so I didn't consider a good idea to use its interface, which should have worked for strings/text.

Below a tenative design:
```python
class BaseTransform:
    def __call__(self, x: str | torch.Tensor ) -> str | torch.Tensor:
        # to be implemented
        pass
    
    def __repr__(self) -> str:
        return f"{self.__class__.__name__}()"
```
* `x` can be either `str`, for transformations that still operate with raw strings or `torch.Tensor`;
* similarly, `__call__:output` can be either `str` or `torch.Tensor`.

Without going into too much detail, but to explain why input and output type may be both... Some transformations we currently have are: (i) `str -> str`, e.g. when we augment data by adding the reverse complement of an amino acid sequence; (ii) others are `str -> torch.Tensor`, e.g. when we need to transform the raw strings into their numerical representation; and (iii) `torch.Tensor -> torch.Tensor`, e.g. when we apply masking.

Tarun-goswamii:
Closes #171

Added trafos.torch module with transforms for DataLoaders.

Includes str->str (Reverse, DNAtoRNA), str->tensor (GreedyEncode), and tensor->tensor (RandomMask) transforms. Also added get_torch_transform() to sklearn base class per discussion in #170.


fkiraly (changes requested):
Can you explain how this design would be used in practice? E.g., by providing a basic vignette on how you expect users to interact with the new classes?

Tarun-goswamii:
@fkiraly plz sir review it .

siddharth7113:
Hi @Tarun-goswamii , I think @fkiraly  meant to give an example in conversation itself and not put it in code right now, I would suggest to revert the commit and post here in the conversation instead first and confirm with him.

Tarun-goswamii:
## Practical Design & Usage 

Dear @fkiraly.
 Here's the actual workflow this enables:

### Real-world pipeline

**Direct chaining example:**
```python
from pyaptamer.trafos.torch import GreedyEncode, RandomMask, DNAtoRNA, Reverse

# String transforms can be chained
dna_sequence = "ACGTACGT"
rna_transform = DNAtoRNA()
reverse_transform = Reverse()

rna_seq = rna_transform(dna_sequence)  # "ACGUACGU"
reversed_seq = reverse_transform(rna_seq)  # "UGCAUGCA"

# Encoding to tensor
vocab = {"A": 1, "C": 2, "G": 3, "U": 4}
encoder = GreedyEncode(vocab=vocab, max_len=16)
encoded = encoder(rna_seq)
# Result: tensor([1, 2, 3, 4, 1, 2, 3, 4, 0, 0, 0, 0, 0, 0, 0, 0])

# Apply augmentation
masker = RandomMask(mask_idx=99, mask_rate=0.15)
masked = masker(encoded)
# Some non-padded positions are randomly replaced with 99

@fkiraly @siddharth7113 plz review it

File: pyaptamer/trafos/torch/_string.py [hunk 1]
```diff
@@ -0,0 +1,31 @@
+"""String-to-string transformations."""
+
+from pyaptamer.trafos.torch._base import BaseTorchTransform
+
+
+class Reverse(BaseTorchTransform):
```
siddharth7113: Why does this have a `BaseTorchTransform`? My understanding is if a transform is using torch underneath we need a  torch transformation, see @fkiraly  comment's [here](https://github.com/gc-os-ai/pyaptamer/issues/171#issuecomment-3389388881)

siddharth7113:
Also please ensure code quality checks are passing in CI, and pre-commit is installed

pyaptamer/trafos/torch/_string.py
```diff
@@ -0,0 +1,21 @@
+"""String-to-string transformations."""
```
siddharth7113: this still doesn't make sense, why have a _string in torch ? when there are no tensor operations involved.

pyaptamer/trafos/__init__.py
```diff
@@ -1 +1 @@
-"""Transformations."""
+"""Transformations module."""
```
siddharth7113: Why add module?

siddharth7113 (changes requested):

pyaptamer/trafos/__init__.py
Tarun-goswamii: thank u , i will correct it!

pyaptamer/trafos/base/_base.py
```diff
@@ -171,3 +171,7 @@ def _check_X(self, X):  # noqa: N802
         """
         X, _ = self._check_X_y(X, None)
         return X
+
+    def get_torch_transform(self):
```
siddharth7113, line 175: What is the reasoning and use of the function that is being added here? Please explain, since it's just returning None?

pyaptamer/trafos/torch/_base.py
```diff
@@ -0,0 +1,19 @@
+"""Base class for PyTorch-compatible transformations."""
+
+from abc import ABC, abstractmethod
+
+from torch import Tensor
+
+
+class BaseTorchTransform(ABC):
```
siddharth7113, line 8: cc @fkiraly , This is the proposed design for the torch base class, but the currently it only puts call and repr, do we have something similar in `sktime` to borrow for torch related transformers?

siddharth7113 (changes requested):
Hi @Tarun-goswamii ,

There are some design decision that would be needed for this PR, we can wait for @fkiraly  response for some time  then raise in discord if needed

pyaptamer/trafos/base/_base.py
Tarun-goswamii: Dear @siddharth7113 The `get_torch_transform()` is designed as an extension point in the sklearn base class. 

By default it returns None, meaning "this transform doesn't have a torch equivalent."

Subclasses that have torch-compatible versions can override this to return them. For example, the sklearn GreedyEncode encoder overrides this method to return the torch GreedyEncode transform with the same parameters.

This allows flexibility - not every sklearn transform needs a torch version, but those that do can provide it through this interface.

PR #251, opened 2026-03-14
fix(aptanet): check pipeline consistency for non deterministic estimators 

### LINKED ISSUE #250: check_pipeline_consistency Fails for Non-Deterministic Estimators
The  `parametrize_with_checks` is used to validate the compatibility of `AptaNetClassifier` and `AptaNetRegressor`. But, the `check_pipeline_consistency` test fails for these estimators because they are non-deterministic.

**Steps to Reproduce**
- Clone the repository and set up the environment.
- Run the following command to execute the test file:
[  python -m pytest pyaptamer/aptanet/tests/test_aptanet.py -v](url)

FrankOHara43:
Resolves #250.

### **Description**

The `parametrize_with_checks` is used to validate the compatibility of `AptaNetClassifier` and `AptaNetRegressor`. But, the `check_pipeline_consistency` test fails for these estimators because they are non-deterministic.

### **Changes**
- **In test_aptanet.py file** 
   - Added a function to mark the `check_pipeline_consistency` test as expected to fail for non-deterministic estimators.
  - Updated `parametrize_with_checks` by replacing manual `if` block workaround with the `expected_failed_checks` .
  - Added `test_expected_failed_checks_marks_non_deterministic_estimators_only` to verify that only non-deterministic estimators are marked as `xfail` for `check_pipeline_consistency`.
- **In  _feature_classifier.py**
   - Added `tags.non_deterministic = True`  to mark the regressor as non-deterministic.

### **How to Test**
- Run this command in the project root: 
     python -m pytest pyaptamer/aptanet/tests/test_aptanet.py -v

### **Output**
- All the checks passed and the non-deterministic ones were skipped 
 
image
image

### DIFF
--- a/pyaptamer/aptanet/tests/test_aptanet.py
+++ b/pyaptamer/aptanet/tests/test_aptanet.py
@@ -84,7 +84,9 @@ def test_pipeline_fit_and_predict_regression(aptamer_seq, protein_seq):
 
 @parametrize_with_checks(
     estimators=[AptaNetClassifier(), AptaNetRegressor()],
-    expected_failed_checks=_expected_failed_checks_for_non_deterministic_estimators,
+    expected_failed_checks={
+        "check_pipeline_consistency": "estimator is non-deterministic"
+        },
 )
 def test_sklearn_compatible_estimator(estimator, check):
     """
@@ -93,22 +95,3 @@ def test_sklearn_compatible_estimator(estimator, check):
     check(estimator)
 
 
-def test_expected_failed_checks_marks_non_deterministic_estimators_only():
-    """
-    Test if pipeline consistency is marked as xfail only for non-deterministic
-    estimators.
-    """
-    expected_failure = {"check_pipeline_consistency": "estimator is non-deterministic"}
-
-    assert (
-        _expected_failed_checks_for_non_deterministic_estimators(AptaNetClassifier())
-        == expected_failure
-    )
-    assert (
-        _expected_failed_checks_for_non_deterministic_estimators(AptaNetRegressor())
-        == expected_failure
-    )
-    assert (
-        _expected_failed_checks_for_non_deterministic_estimators(LinearRegression())
-        == {}
-    )
--- a/pyaptamer/aptanet/tests/test_aptanet.py
+++ b/pyaptamer/aptanet/tests/test_aptanet.py
@@ -3,7 +3,6 @@ __author__ = ["nennomp", "satvshr"]
 
 import numpy as np
 import pytest
-from sklearn.linear_model import LinearRegression
 from sklearn.utils.estimator_checks import parametrize_with_checks
 
 from pyaptamer.aptanet import AptaNetClassifier, AptaNetPipeline, AptaNetRegressor
@@ -86,12 +85,10 @@ def test_pipeline_fit_and_predict_regression(aptamer_seq, protein_seq):
     estimators=[AptaNetClassifier(), AptaNetRegressor()],
     expected_failed_checks={
         "check_pipeline_consistency": "estimator is non-deterministic"
-        },
+    },
 )
 def test_sklearn_compatible_estimator(estimator, check):
     """
     Run scikit-learn's compatibility checks on the AptaNetClassifier.
     """
     check(estimator)
-
-
--- a/pyaptamer/aptanet/tests/test_aptanet.py
+++ b/pyaptamer/aptanet/tests/test_aptanet.py
@@ -15,17 +15,6 @@ params = [
 ]
 
 
-def _expected_failed_checks_for_non_deterministic_estimators(estimator):
-    """
-    Mark pipeline consistency check is expected to fail only for non-deterministic
-    estimators.
-    """
-    sklearn_tags = estimator.__sklearn_tags__()
-    if sklearn_tags.non_deterministic:
-        return {"check_pipeline_consistency": "estimator is non-deterministic"}
-    return {}
-
-
 @pytest.mark.parametrize("aptamer_seq, protein_seq", params)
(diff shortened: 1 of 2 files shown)


FrankOHara43:
fixes #250

satvshr:
@NennoMP this is for you given it was your TODO

pyaptamer/aptanet/tests/test_aptanet.py
```diff
@@ -15,6 +16,17 @@
...
+    """
+    Mark pipeline consistency check is expected to fail only for non-deterministic
+    estimators.
+    """
+    sklearn_tags = estimator.__sklearn_tags__()
+    if sklearn_tags.non_deterministic:
+        return {"check_pipeline_consistency": "estimator is non-deterministic"}
+    return {}
+
+
```
NennoMP: Isn't this essentially the same as what we had commented out?

```python
# expected_failed_checks={
#    "check_pipeline_consistency": "estimator is non-deterministic"
# },
```

I also think we do not really need to check for non-determinism since `AptaNet` (both the classifier and regressor) will always be non-deterministic.

pyaptamer/aptanet/tests/test_aptanet.py
```diff
@@ -15,6 +16,17 @@
...
+def _expected_failed_checks_for_non_deterministic_estimators(estimator):
+    """
+    Mark pipeline consistency check is expected to fail only for non-deterministic
+    estimators.
+    """
+    sklearn_tags = estimator.__sklearn_tags__()
+    if sklearn_tags.non_deterministic:
+        return {"check_pipeline_consistency": "estimator is non-deterministic"}
+    return {}
+
```
NennoMP: Isn't this essentially the same as what we had commented out?

```python
# expected_failed_checks={
#    "check_pipeline_consistency": "estimator is non-deterministic"
# },
```

Also, I think we don't need to check for non-determinism with an `if` statement, since `AptaNet` (both the classifier and regressor) will always be non-deterministic by design.

NennoMP (commented):
I just checked whether the following code works the same or not. Locally all tests are passing, including `check_pipeline_consistency`.

```python
@parametrize_with_checks(
    estimators=[AptaNetClassifier(), AptaNetRegressor()],
    # I uncommented the code below
    expected_failed_checks={
        "check_pipeline_consistency": "estimator is non-deterministic"
    },
)
def test_sklearn_compatible_estimator(estimator, check):
    """
    Run scikit-learn's compatibility checks on the AptaNetClassifier.
    """
    # I removed the `if-raise` statement
    check(estimator)
```

Can you verify this @FrankOHara43? Obviously, we still need to re-add `tags.non_deterministic = True`, which you already did.

If the code above works, I would adopt such a design since it's way simpler that the current PR. I don't think we really need the method `_expected_failed_checks_for_non_deterministic_estimators`. Then we can merge.

I think that, back when we had this issue, the test was failing due to a bug on `sklearn` side, which was still executing the pipeline consistency check despite it being expected to fail.

pyaptamer/aptanet/tests/test_aptanet.py
```diff
@@ -72,23 +84,31 @@ def test_pipeline_fit_and_predict_regression(aptamer_seq, protein_seq):
...
+        == expected_failure
+    )
+    assert (
+        _expected_failed_checks_for_non_deterministic_estimators(AptaNetRegressor())
+        == expected_failure
+    )
+    assert (
+        _expected_failed_checks_for_non_deterministic_estimators(LinearRegression())
+        == {}
+    )
```
NennoMP: See my review main comment. If that works can we also remove this? Not sure what it does or if that's strictly needed.

FrankOHara43:
@NennoMP I applied the changes that were suggested in the above comments. One check seems to fail though.
image
 I applied this: 
 1. I used the code you gave in the above main comment

   ```
 @parametrize_with_checks(
    estimators=[AptaNetClassifier(), AptaNetRegressor()],
    # I uncommented the code below
    expected_failed_checks={
        "check_pipeline_consistency": "estimator is non-deterministic"
    },
)
def test_sklearn_compatible_estimator(estimator, check):
    """
    Run scikit-learn's compatibility checks on the AptaNetClassifier.
    """
    # I removed the `if-raise` statement
    check(estimator) 
```
 2.  and I kept `tags.non_deterministic = True`  and reverted rest of the changes.

NennoMP:
@FrankOHara43 The test which fails seems to be `check_estimators_pickle(readonly_memmap=True)`, correct? 

However, I am not sure whether the issue is within the code, in your screenshot the error message looks like this `OS Error: Too many files open`.

When I run these tests with the code I propose I see all checks passing. Can you push the new changes so that we may check whether tests are passing with CI/CD? Worst case scenario we revert to what we have now.

FrankOHara43:
@NennoMP Thank you. Yes, the failing test is indeed `check_estimators_pickle(readonly_memmap=True)` . It seems to be caused by a local file descriptor limit . I've pushed the changed as suggest above please take a look.

NennoMP:
> @NennoMP Yes, the failing test is indeed `check_estimators_pickle(readonly_memmap=True)` . Please take a look at the changes made.

Please make sure to run pre-commit (e.g., `pre-commit run --all-files`), code quality is failing. Otherwise tests are not even executed.

FrankOHara43:
@NennoMP  I increased the limit of my local file descriptor and all checks seem to pass 
image

NennoMP:
The originally commented-out approach seems to work fine.

Remove all unnecessary changes (method `_expected_failed_checks_for_non_deterministic_estimators`) and I'll merge this.

FrankOHara43:
@NennoMP removed all the unnecessary changes as suggested in the above comment.

PR #252, opened 2026-03-15
Refactor: Improve robustness and readability of InteractionMap forward pass

kamilansri:
## Description
This PR refactors the `InteractionMap` module to improve its robustness and readability during the forward pass computations. 

## Changes Made
* **Robust Shape Validation:** Replaced the `assert` statement checking for matching feature dimensions with an explicit `ValueError`. This ensures the validation step is strictly enforced and isn't silently ignored if the environment runs Python with optimization flags (`-O`).
* **Idiomatic Tensor Operations:** Updated the dot product computation, replacing `torch.matmul` and tensor transposition with `torch.einsum('bid,bjd->bij', x_apta, x_prot)`. This makes the intended axes of multiplication explicitly clear and aligns with PyTorch best practices.

## Impact
* Zero impact on mathematical output or the module's public API.
* Improved code maintainability and safer runtime validation.

### DIFF
--- a/pyaptamer/aptatrans/layers/_interaction_map.py
+++ b/pyaptamer/aptatrans/layers/_interaction_map.py
@@ -32,21 +32,24 @@ class InteractionMap(nn.Module):
         ----------
         x_apta, x_prot : Tensor
             Input tensor for aptamers and proteins, respectively. Shapes are
-            (batch_size, seq_len (s1), n_features) and (batch_size, seq_len (s2),
+            (batch_size, seq_len_1, n_features) and (batch_size, seq_len_2,
             n_features), respectively.
 
         Returns
         ----------
         Tensor
-            Interaction map tensor of shape (batch_size, 1, seq_len (s1), seq_len (s2)).
+            Interaction map tensor of shape (batch_size, 1, seq_len_1, seq_len_2).
         """
-        assert x_apta.shape[-1] == x_prot.shape[-1], (
-            "The number of features of `x_apta` and `x_prot` must match."
-        )
+        if x_apta.shape[-1] != x_prot.shape[-1]:
+            raise ValueError(
+                f"Feature dimensions must match. Got {x_apta.shape[-1]} and {x_prot.shape[-1]}."
+            )
 
-        # compute interaction matrix using batch matrix multiplication
-        out = torch.matmul(x_apta, x_prot.transpose(-2, -1))  # (batch_size, s1, s2)
-        # add channel dimension for compatibility with nn.Conv2d operations
-        out = out.unsqueeze(1)  # (batch_size, s1, s2) -> (batch_size, 1, s1, s2)
+        # Compute interaction matrix using Einstein summation
+        # b = batch_size, i = seq_len_1, j = seq_len_2, d = n_features
+        out = torch.einsum('bid,bjd->bij', x_apta, x_prot) 
+
+        # Add channel dimension for compatibility with nn.BatchNorm2d
+        out = out.unsqueeze(1)  # (batch_size, 1, s1, s2)
 
         return self.batchnorm(out)


pyaptamer/aptatrans/layers/_interaction_map.py
```diff
@@ -32,21 +32,24 @@ def forward(self, x_apta: Tensor, x_prot: Tensor) -> Tensor:
         ----------
         x_apta, x_prot : Tensor
             Input tensor for aptamers and proteins, respectively. Shapes are
-            (batch_size, seq_len (s1), n_features) and (batch_size, seq_len (s2),
+            (batch_size, seq_len_1, n_features) and (batch_size, seq_len_2,
```
NennoMP, line 35: Why are we changing docstrings when not needed (since logic stayed the same)?

pyaptamer/aptatrans/layers/_interaction_map.py
```diff
@@ -32,21 +32,24 @@ def forward(self, x_apta: Tensor, x_prot: Tensor) -> Tensor:
...
             n_features), respectively.
 
         Returns
         ----------
         Tensor
-            Interaction map tensor of shape (batch_size, 1, seq_len (s1), seq_len (s2)).
+            Interaction map tensor of shape (batch_size, 1, seq_len_1, seq_len_2).
```
NennoMP, line 41: See above.

pyaptamer/aptatrans/layers/_interaction_map.py
```diff
@@ -32,21 +32,24 @@ def forward(self, x_apta: Tensor, x_prot: Tensor) -> Tensor:
...
-        )
-
-        # compute interaction matrix using batch matrix multiplication
-        out = torch.matmul(x_apta, x_prot.transpose(-2, -1))  # (batch_size, s1, s2)
-        # add channel dimension for compatibility with nn.Conv2d operations
-        out = out.unsqueeze(1)  # (batch_size, s1, s2) -> (batch_size, 1, s1, s2)
+        if x_apta.shape[-1] != x_prot.shape[-1]:
+            raise ValueError(
+                f"Feature dimensions must match. Got {x_apta.shape[-1]} and {x_prot.shape[-1]}."
+            )
```
NennoMP, line 46: Here, we are using an `assert` as an executable comment, rather than validation. The goal is to tell the reader the assumption we make about `x_apta` and `x_prot` shapes. 

By the time we reach this point in the AptaTrans pipeline the caller should have already ensured that dimensions are compatible. I don't see any reason to raise an exception here.

pyaptamer/aptatrans/layers/_interaction_map.py
```diff
@@ -32,21 +32,24 @@ def forward(self, x_apta: Tensor, x_prot: Tensor) -> Tensor:
...
+
+        # Compute interaction matrix using Einstein summation
+        # b = batch_size, i = seq_len_1, j = seq_len_2, d = n_features
+        out = torch.einsum('bid,bjd->bij', x_apta, x_prot) 
+
+        # Add channel dimension for compatibility with nn.BatchNorm2d
+        out = out.unsqueeze(1)  # (batch_size, 1, s1, s2)
```
NennoMP, line 53: This change is also not needed... you are simply replacing `torch.matmul` with an equivalent operation in `torch.einsum`.

If anything, I believe that the former version with `torch.matmul` is more readable and familiar for PyTorch users; rather than the opposite.

NennoMP:
My review comments are above.

Why are we changing this? As you also mentioned the logic is the same as before and we simply changed two things: (i) docstring format for not apparent reason and (ii) replaced `torch.matmul` with an equivalent operation leveraging `torch.einsum`.


Overall, I think this is an unnecessary change. Also, we didn't actually have a corresponding issue for this PR. If you'd like to contribute to `pyaptamer` I would suggest to focus on pre-existing issue or at least to create one before starting a PR, so that maintainers can discuss whether the issue is relevant or not.

kamilansri:
Thanks @NennoMP ,
I appreciate your review, I will take care about the same next time.

PR #259, opened 2026-03-19
Adding different formats to molecule loader

### LINKED ISSUE #202: [ENH] Extend `MoleculeLoader`

siddharth7113:
Resolves #202 

Adapted from #205 to the new output format from #203


pyaptamer/data/loader.py
```diff
@@ -12,58 +12,75 @@
...
-        the following formats are currently supported:
+    Supported file types
+    --------------------
+    The loader determines the type from the file extension by default, but a
+    format override can be provided to force a specific loader:
 
-        * ``pdb`` format
+    - ``fmt="pdb"`` -> handled by :func:`pdb_to_aaseq` (PDB SEQRES extraction)
+    - any other format string (e.g. ``"fasta"``, ``"genbank"``) -> passed to
+      :mod:`Bio.SeqIO` for parsing
```
satvshr, line 37: Seems unnecessary if we are mentioning "all formats supported by :func:`Bio.SeqIO.parse` are accepted", though it is good to mention pdb along with this given `Bio.SeqIO.parse` does not support pdb as far as I know?

pyaptamer/data/loader.py
```diff
@@ -74,16 +91,51 @@ def to_df_seq(self):
...
+            Format string such as "pdb", "fasta", "genbank", etc.
+        """
+        # If user provided a format override, use it
+        if self.fmt is not None:
+            return self.fmt
+
+        # Fallback to suffix-based inference
         suffix = path.suffix.lower()
         if suffix == ".pdb":
             return "pdb"
```
satvshr: Seems unnecessary given `return suffix.lstrip(".") if suffix else None`

pyaptamer/utils/tests/test_hf_to_dataset.py
```diff
@@ -0,0 +1,20 @@
+import os
```
satvshr: I assume this PR either stacks on a PR or is not updated with main? This file seems unrelated along with other hf related files like `pyaptamer/utils/__init__.py` and `pyaptamer/utils/_hf_to_dataset.py`

pyaptamer/datasets/tests/test_loaders_mol.py
```diff
@@ -27,3 +29,21 @@ def test_loader_mol_to_df_seq():
     assert df.shape == (10, 1)
     seq = df.iloc[0, 0]
     assert seq.startswith("QTDMSRK")
+
+
+def test_loader_seqio_fasta_to_df_seq(tmp_path):
```
satvshr: Could you add a multiindex df test instead? To have a broader and realsitic scope of outcomes covered?

satvshr (commented):
Looks good! Took a quick look and left some comments, will take a deeper look before merging finally

pyaptamer/data/loader.py
siddharth7113: while `SeqIO` supports all formats we are currently passing through a different function which internally uses SeqIO  since we need to an additional step of extracting chanins in pdb format.

pyaptamer/utils/tests/test_hf_to_dataset.py
siddharth7113: lol, I know idea what is it , but it seems this is what you orignally wrote in the #205 , Should I delete it?

pyaptamer/datasets/tests/test_loaders_mol.py
siddharth7113: What do you mean by multiindex df test? Do you mean multiple file in same data format?

satvshr: Multiple files with multiple chains ideally

pyaptamer/utils/tests/test_hf_to_dataset.py
satvshr: That PR has not been updated with main in months now, this is already merged to main as far as I remember

siddharth7113: I am not sure I updated and merged with main.

satvshr: That is because you added this to the utils directory, it is merged [here](https://github.com/gc-os-ai/pyaptamer/blob/main/pyaptamer/datasets/tests/test_hf_to_dataset.py)

siddharth7113:
@satvshr ready for re-review

pyaptamer/data/loader.py
```diff
@@ -12,58 +12,75 @@
...
+    ----------
+    path : str, Path, or list
+        File location(s) of molecule files. One row is returned per file.
+    index : list, or pandas.Index coercible, optional
+        Row index for the resulting DataFrame; if None, integer RangeIndex is used.
     columns : list, optional
         column names for the structure; if None, defaults to ["sequence"]
 
     ignore_duplicates : bool, optional, default=False
         if True, removes duplicate sequences (keeping the first occurrence).
```
satvshr: Can we test each of these parameter inputs?

siddharth7113: good catch the docstring is also wrong, when we use MoleculeLoader it is now mulitindex as default, and currently both ignore_duplicatesd and index were not being used.

pyaptamer/data/loader.py
```diff
@@ -12,78 +12,138 @@
...
 
-    Parameters
-    ------------
-    path : str, Path, or list thereof
-        file location or list of file locations with molecule files
-        str can be any of the following... [fill in]
+    This loader extracts primary amino-acid sequences from molecule files.
 
-        the following formats are currently supported:
+    Supported file types
```
satvshr, line 30: This should come after parameters I think

pyaptamer/data/loader.py
```diff
@@ -12,78 +12,138 @@
...
             self._path = [Path(p) if isinstance(p, str) else p for p in path]
+        else:
+            raise TypeError("path must be a str, Path, or list of str/Path")
 
     def to_df_seq(self):
         """Return a pd.DataFrame of sequences with MultiIndex (path, chain_id).
 
         Returns
-        --------
+        -------
```
satvshr: Kindly revert, unless ruff is making these changes

pyaptamer/data/loader.py
satvshr: Tests are still not covering every possible parameter set? I meant if there are 4 parameters and each has say 3 possible values, then all 12 combinations should be tested.

siddharth7113: That would be 81 cases, I don't think it make sense to cover all possible cases.

siddharth7113: I think the best we can do is pairwise coverage for meangiful cases.

satvshr (approved):
Nice! Didn't check the additional tests, but based on conversations on Discord (and the urgency of this) I will assume we are covering as much as possible given this is something that will be used everywhere.

Issue #230, opened 2026-02-16
[BUG] Finetuning crashes with AMP: `binary_cross_entropy` is unsafe to autocast (needs BCEWithLogitsLoss or disable autocast for loss)


siddharth7113:
### Summary
When running the finetuning stage of AptaTrans with Lightning `precision=16-mixed` (AMP), training crashes immediately with:

> `RuntimeError: torch.nn.functional.binary_cross_entropy and torch.nn.BCELoss are unsafe to autocast.`

This happens inside `pyaptamer/aptatrans/_model_lightning.py` during finetune loss computation.

Pretraining runs fine (both apta and prot encoders). The crash only appears in finetuning.

### Error / Traceback
(Attaching full log as file as well)

```text
Traceback (most recent call last):
0: [rank0]:   File "/home/siddharth/work/pyaptamer/train_aptatrans.py", line 535, in <module>
0: [rank0]:     main()
0: [rank0]:     ~~~~^^
0: [rank0]:   File "/home/siddharth/work/pyaptamer/train_aptatrans.py", line 526, in main
0: [rank0]:     trainer.fit(lit, train_dataloaders=train_dl, ckpt_path=ckpt_path)
0: [rank0]:     ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
0: [rank0]:   File "/home/siddharth/work/pyaptamer/.venv/lib/python3.13/site-packages/lightning/pytorch/trainer/trainer.py", line 584, in fit
0: [rank0]:     call._call_and_handle_interrupt(
0: [rank0]:     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
0: [rank0]:         self,
0: [rank0]:         ^^^^^
0: [rank0]:     ...<6 lines>...
0: [rank0]:         weights_only,
0: [rank0]:         ^^^^^^^^^^^^^
0: [rank0]:     )
0: [rank0]:     ^
0: [rank0]:   File "/home/siddharth/work/pyaptamer/.venv/lib/python3.13/site-packages/lightning/pytorch/trainer/call.py", line 48, in _call_and_handle_interrupt
0: [rank0]:     return trainer.strategy.launcher.launch(trainer_fn, *args, trainer=trainer, **kwargs)
[...]

PR #265, opened 2026-03-31
Fix AMP training stability by switching to BCEWithLogitsLoss (#230)

hyndhavamahesh345:
**Problem:** Training currently crashes when using Automatic Mixed Precision (AMP) because F.binary_cross_entropy is unsafe to autocast and requires inputs to be strictly in the range [0, 1].

**Solution:**

Model Refactor: Removed nn.Sigmoid from the AptaTrans model's final head. Outputting raw logits is the standard practice for modern deep learning.
Loss Stability: Switched to F.binary_cross_entropy_with_logits in the Lightning module. This is numerically more stable and explicitly supports AMP/autocast.
Backward Compatibility: Updated AptamerEvalAptaTrans.evaluate to explicitly apply a Sigmoid activation, ensuring that affinity scores returned to users remain as probabilities (0 to 1).
Accuracy Update: Updated accuracy logic to handle logits before thresholding.

**Testing:**

Updated test_aptatrans.py to account for the model change.
Verified that all 34 core unit tests and 7 experimental tests pass successfully.

### DIFF
--- a/pyaptamer/aptatrans/_model.py
+++ b/pyaptamer/aptatrans/_model.py
@@ -162,7 +162,6 @@ class AptaTrans(nn.Module):
                     ("linear1", nn.Linear(self.inplanes, self.inplanes // 2)),
                     ("activation1", nn.GELU()),
                     ("linear2", nn.Linear(self.inplanes // 2, 1)),
-                    ("activation2", nn.Sigmoid()),
                 ]
             )
         )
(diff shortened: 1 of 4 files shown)


siddharth7113:
Please see [here](https://github.com/gc-os-ai/pyaptamer/issues/230#issuecomment-4204343391)

Issue #260, opened 2026-03-21
Optimize CI workflow by removing redundant checkout step and improving dependency caching


ruchita-mlcraft:
# Description
While reviewing the CI workflow defined in  [github/workflows/test.yml] observed a redundancy and a potential performance optimization opportunity.

#Observations
1. Duplicate Checkout Step
   The **code-quality** job includes two **actions/checkout** steps:
   - One initial checkout
   - Another with **fetch-depth: 0**
  This leads to redundant repository fetching within the same job.
2. No Dependency Caching
  - The workflow installs dependencies using pip in multiple jobs, but currently does not utilize caching. This increases CI execution time
    unnecessarily.

# Proposed Improvements
- Remove the redundant initial checkout step and retain the one with **fetch-depth: 0**
- Introduce pip caching using **actions/cache** to speed up dependency installation across runs

# Expected Impact
- Reduced CI execution time
- Improved workflow efficiency
- Cleaner and more maintainable CI configuration

# Additional Context
This improvement does not affect application logic and is limited to CI optimization.

# Contribution
I would like to work on this issue and submit a PR implementing the above improvements. Please let me know if this approach aligns with the project guidelines.

PR #266, opened 2026-03-31
Optimize CI workflow and enable caching (#260)

hyndhavamahesh345:
**Problem:** 

The CI pipeline was inefficient due to redundant repository checkout steps and the lack of package caching, leading to unnecessarily long build times.

**Solution:**

**Streamlined Checkout:** Removed a duplicate actions/checkout@v6 step in the code-quality job and consolidated it with required fetch-depth settings.

**Dependency Caching:** Added cache: "pip" and cache-dependency-path: "pyproject.toml" to all setup-python jobs. This allows GitHub Actions to reuse installed packages across different runs.

**Impact:**

Significantly reduces runtime for CI checks by skipping the download phase for unchanged dependencies.
Reduces resource consumption on CI runners.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,19 +27,18 @@ jobs:
     steps:
       - name: repository checkout step
         uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       - name: python environment step
         uses: actions/setup-python@v6
         with:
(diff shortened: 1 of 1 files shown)


satvshr:
Share a benchmark comparing this v/s current system to show runtime has decreased, thanks

PR #267, opened 2026-03-31
Prepare release 0.1.0a1

siddharth7113:


README.md
```diff
@@ -32,15 +32,14 @@ Checkout [examples/](examples) to see how to use the current API.
...
-### Stable version from pip
-
-NOTE: the package is not released yet on pypi - install from repo instead (unstable version)
+### PyPI prerelease
 
 ```bash
-pip install pyaptamer
+pip install pyaptamer==0.1.0a1
 ```
+NOTE: pyaptamer is in early development. The API is unstable and may change between releases.
```
satvshr, line 40: Do we really have to mention the unstable API again? It's mentioned in line 23

siddharth7113:
Interesting found a bug during release.

siddharth7113:
Failing due to unrelated reasons. 
The function [here](https://github.com/gc-os-ai/pyaptamer/blob/main/pyaptamer/utils/tests/test_pdb_to_seq_uniprot.py) assumes server always return df, but it fails when server is done. I would open an issue for the same.

README.md
siddharth7113: We can change that but I honestly see no harm, I want our users to know we will change the API 😂.

jk, I will update it, I think part of it was repeated from old readme.

Issue #284, opened 2026-04-04
[BUG] `rna2vec` dropping shorter sequences <3 without correct error


Rusheel86:
**Is your feature request related to a problem? Please describe.**
The core models, specifically `AptaTrans` and `AptaNetMLP`, currently lack defensive checks at the start of their `forward` passes. This makes the library vulnerable to cryptic downstream errors when messy biological data is processed.
For example, if an upstream data loader or masking logic produces an empty tensor (shape `[batch, 0]`), the error isn't caught until deep inside the C++ backend of a convolutional layer or batch norm, resulting in a confusing `RuntimeError: Calculated padded input size per channel: (0 x 0)` that obscures the true cause.

**Describe the solution you'd like**
I propose adding a centralized, lightweight validation utility (potentially in `src/pyaptamer/utils/_validation.py`) that checks for:

- Empty Tensors: Ensuring `.numel() > 0` before entering encoders.
- NaN/Inf Detection: Catching numerical instabilities early.
- Shape Consistency: Basic checks for sequence length vs. expected dimensions.

This utility would be injected into the `forward` methods of:

- AptaTrans (in `_model.py`)
- InteractionMap (in `layers/_interaction_map.py`)
- AptaNetMLP (in `_aptanet_nn.py`)

**Describe alternatives you've considered**
I considered adding manual if checks directly into each forward method, but a centralized utility ensures a consistent error message format across the entire library and is easier to maintain as new architectures (like AptaTrans variants) are added.

**Additional context**
Adding these checks will significantly improve the developer experience for researchers using pyaptamer for in-silico aptamer design, where data edge cases (short sequences, missing protein structures) are common.

I have a local environment set up and would be happy to submit a PR for this if the approach looks good :)

siddharth7113:
> This makes the library vulnerable to cryptic downstream errors when messy biological data is processed.
For example, if an upstream data loader or masking logic produces an empty tensor (shape [batch, 0]), the error isn't caught until deep inside the C++ backend of a convolutional layer or batch norm, resulting in a confusing RuntimeError: Calculated padded input size per channel: (0 x 0) that obscures the true cause.



Hi, Can you please describe a code-block or case where this would happen, I am unable to understand where this would happen in currently exposed public-API?

Lakshya-2440:
Yes, the issue is valid overall, with one nuance.
I reproduced this via the exposed pipeline and got:
[...]

PR #288, opened 2026-04-04
Fix rna2vec silently dropping short sequences

WHOIM1205:
### Fix: preserve row alignment in `rna2vec` to prevent dataset corruption

**Summary**  
Fixes a critical bug where `rna2vec` silently dropped sequences (e.g., short or invalid ones), causing misalignment between aptamer, protein, and label arrays during training.

**Changes**
- Removed conditional skip (`if any(converted)`) in `rna2vec`
- Ensured every input sequence produces an output row (zero-padded if needed)
- Updated edge-case tests to reflect new behavior
- Added invariant test to guarantee `len(output) == len(input)`

**Impact**
- Prevents silent data corruption in training datasets  
- Fixes incorrect aptamer–protein–label pairing  
- Avoids crashes during evaluation on short/invalid sequences  
- Maintains backward compatibility for valid inputs  

**Note**  
This restores the expected 1:1 mapping between input sequences and encoded outputs, which downstream components rely on.

### DIFF
--- a/pyaptamer/utils/_rna.py
+++ b/pyaptamer/utils/_rna.py
@@ -151,24 +151,22 @@ def rna2vec(
             triplets.get(sequence[i : i + 3], 0) for i in range(len(sequence) - 2)
         ]
 
-        # skip sequences that convert to an empty list
-        if any(converted):
-            # truncate if too long
-            if max_sequence_length is not None and len(converted) > max_sequence_length:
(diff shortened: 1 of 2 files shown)


siddharth7113:
See [here](https://github.com/gc-os-ai/pyaptamer/issues/284#issuecomment-4281195043)

PR #294, opened 2026-04-05
Reconstruct encoded sequence before evaluation in MCTS simulation

WHOIM1205:
### What was happening
During simulation, we were passing encoded sequences (with `_` markers) directly to the evaluator. These got converted into invalid sequences internally (full of `N`s), so the scores driving the MCTS search were basically meaningless.

### What’s changed
We now reconstruct the sequence using `_reconstruct()` before evaluation, so the model sees the actual nucleotide sequence.

### Why it matters
This ensures the search is guided by real scores instead of corrupted ones, making the generated candidates much more reliable.


NennoMP:
@WHOIM1205 You are right. Thanks for spotting this!

PR #299, opened 2026-04-05
[BUG] Fix Benchmarking duplicate estimator label collisions

### LINKED ISSUE #297: [BUG] Benchmarking collapses multiple estimators of the same class into one result row
## Describe the bug
`Benchmarking.run()` stores results in a dict keyed only by `estimator.__class__.__name__`, so benchmarking two estimators of the same class silently overwrites the first result with the second.

This means hyperparameter comparisons within the same estimator family are not represented correctly in the summary table.

Relevant code:
- `pyaptamer/benchmarking/_base.py` (`est_name = estimator.__class__.__name__` and `results[est_name] = est_scores`)

## Minimal reproduction
```python
import numpy as np
from sklearn.dummy import DummyClassifier
from sklearn.metrics import accuracy_score
from pyaptamer.benchmarking import Benchmarking

X = np.array([[0], [1], [0], [1], [0], [1], [0], [1]])
y = np.array([0, 1, 0, 1, 0, 1, 0, 1])

bench = Benchmarking(
    estimators=[
        DummyClassifier(strategy="most_frequent"),
        DummyClassifier(strategy="stratified", random_state=0),
    ],
    metrics=[accuracy_score],
    X=X,
    y=y,
    cv=2,
)

print(bench.run())
```

Current behavior:
- only one `DummyClassifier` row is returned

Expected behavior:
- both estimators should appear in the results under distinct labels
- distinct estimator classes should keep their original class-name labels

## Proposed fix
Generate stable display names for estimators before running CV, and disambiguate duplicate class names with a deterministic suffix such as `[1]`, `[2]`, etc.

ashum9:
Closes #297.

## What this changes
- disambiguates duplicate estimator class names in `Benchmarking` results with deterministic 1-based suffixes
- preserves the original class name when there is only one estimator of that class
- adds a regression test proving two `DummyClassifier` instances no longer overwrite one another
- adds a regression test proving distinct estimator classes keep their original labels

## Validation
- `pytest -q pyaptamer/benchmarking/tests/test_benchmarking_core.py`

### DIFF
--- a/pyaptamer/benchmarking/tests/test_benchmarking_core.py
+++ b/pyaptamer/benchmarking/tests/test_benchmarking_core.py
@@ -23,8 +23,8 @@ def test_benchmarking_keeps_duplicate_estimator_classes_distinct():
 
     summary = bench.run()
 
-    assert ("DummyClassifier[1]", "accuracy_score") in summary.index
-    assert ("DummyClassifier[2]", "accuracy_score") in summary.index
+    assert ("DummyClassifier_1", "accuracy_score") in summary.index
+    assert ("DummyClassifier_2", "accuracy_score") in summary.index
     assert len(summary) == 2
 
 
--- a/pyaptamer/benchmarking/tests/test_benchmarking_core.py
+++ b/pyaptamer/benchmarking/tests/test_benchmarking_core.py
@@ -1,0 +1,46 @@
+import numpy as np
+from sklearn.dummy import DummyClassifier, DummyRegressor
+from sklearn.metrics import accuracy_score
+
+from pyaptamer.benchmarking._base import Benchmarking
+
+
+def test_benchmarking_keeps_duplicate_estimator_classes_distinct():
+    """Estimators with the same class should not overwrite one another."""
+    X = np.array([[0], [1], [0], [1], [0], [1], [0], [1]])
+    y = np.array([0, 1, 0, 1, 0, 1, 0, 1])
+
+    bench = Benchmarking(
+        estimators=[
+            DummyClassifier(strategy="most_frequent"),
+            DummyClassifier(strategy="stratified", random_state=0),
+        ],
+        metrics=[accuracy_score],
+        X=X,
+        y=y,
+        cv=2,
+    )
+
+    summary = bench.run()
+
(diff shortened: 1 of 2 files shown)


pyaptamer/benchmarking/tests/test_benchmarking_core.py
```diff
@@ -0,0 +1,46 @@
...
+        ],
+        metrics=[accuracy_score],
+        X=X,
+        y=y,
+        cv=2,
+    )
+
+    summary = bench.run()
+
+    assert ("DummyClassifier[1]", "accuracy_score") in summary.index
```
siddharth7113: this notation wouldn't be right, and could be confusing in downstream usage, user would have to make a call something like this `df.loc["DummyClassifier[1]"]` , I would suggest to index along with strategy something to make it either like `DummyClassifer_1` or based on Pattern should be fine as well.

siddharth7113 (changes requested):
left the comment to change the suffix structure, otherwise looks good.

ashum9:
@siddharth7113 Thanks for the feedback. I updated the duplicate-estimator label format to use underscore suffixes instead of brackets.

Now duplicate estimators of the same class are labeled deterministically as `DummyClassifier_1`, `DummyClassifier_2` (1-based, only when needed), while unique estimator classes keep their original class name.

Updated the regression tests accordingly; `pytest -q pyaptamer/benchmarking/tests/test_benchmarking_core.py` passes.

siddharth7113:
Please follow PR guidelines and make sure to install and run `pre-commit` , the current CI checks are failing.

siddharth7113:
cc @satvshr , for your feedback

pyaptamer/benchmarking/tests/test_benchmarking_core.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    summary = bench.run()
+
+    assert ("DummyClassifier_1", "accuracy_score") in summary.index
+    assert ("DummyClassifier_2", "accuracy_score") in summary.index
+    assert len(summary) == 2
+
+
+def test_benchmarking_preserves_unique_estimator_names():
```
satvshr, line 31: Was this also a problem which you solved or are you adding a random test

PR #300, opened 2026-04-05
[BUG] Handle missing package metadata when importing from source

### LINKED ISSUE #298: [BUG] Importing pyaptamer from a source checkout fails without installed package metadata
## Describe the bug
Importing `pyaptamer` from a plain source checkout fails when the distribution metadata is not installed.

Relevant code:
- `pyaptamer/__init__.py` calls `version("pyaptamer")` unconditionally

## Reproduction
From a fresh checkout, before running `pip install -e .`:
```bash
python -c "import pyaptamer"
```

Current behavior:
- raises `importlib.metadata.PackageNotFoundError: No package metadata was found for pyaptamer`

Expected behavior:
- `import pyaptamer` should succeed from a checkout
- `__version__` should fall back gracefully when package metadata is unavailable

## Why this matters
This breaks local development workflows and causes test collection/imports to fail in environments that import the package directly from source.

## Proposed fix
Catch `PackageNotFoundError` in `pyaptamer/__init__.py` and provide a safe fallback version string when metadata is unavailable.

ashum9:
Closes #298.

## Summary
Importing `pyaptamer` directly from a source checkout can fail when distribution metadata is not present (e.g. no editable/wheel install, no `pyaptamer.egg-info`). This PR makes `pyaptamer` import robust in that scenario.

## What changed
- `pyaptamer/__init__.py`: wrap `importlib.metadata.version("pyaptamer")` in `try/except PackageNotFoundError` and fall back to `__version__ = "0+unknown"`.
- `pyaptamer/tests/test_import.py`: add a subprocess regression test that forces the no-metadata path by patching `importlib.metadata.version` to raise `PackageNotFoundError` before importing `pyaptamer`.

## Validation
- `pytest -q pyaptamer/tests/test_import.py`

### DIFF
--- a/pyaptamer/__init__.py
+++ b/pyaptamer/__init__.py
@@ -1,5 +1,8 @@
 """pyaptamer: Python library for aptamer design."""
 
-from importlib.metadata import version
+from importlib.metadata import PackageNotFoundError, version
 
-__version__ = version("pyaptamer")
+try:
+    __version__ = version("pyaptamer")
+except PackageNotFoundError:
+    __version__ = "0+unknown"
--- a/pyaptamer/tests/test_import.py
+++ b/pyaptamer/tests/test_import.py
@@ -1,0 +1,17 @@
+"""Tests for package import behavior."""
+
+import subprocess
+import sys
+
+
+def test_import_from_source_checkout_without_installed_metadata():
+    """Importing pyaptamer from a checkout should not require wheel metadata."""
+    result = subprocess.run(
+        [sys.executable, "-c", "import pyaptamer; print(pyaptamer.__version__)"],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+    assert result.returncode == 0, result.stderr
+    assert result.stdout.strip()
--- a/pyaptamer/tests/test_import.py
+++ b/pyaptamer/tests/test_import.py
@@ -6,12 +6,32 @@ import sys
 
 def test_import_from_source_checkout_without_installed_metadata():
     """Importing pyaptamer from a checkout should not require wheel metadata."""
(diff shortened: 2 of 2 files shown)


siddharth7113:
I am closing this PR, since I am unable to reproduce this issue even with a fresh install, if you think this bug is valid and still exsist, feel free to open an issue and give clear instructions on reproducing the issue.

siddharth7113:
If you are using AI to raise PR, please verify if the issue is valid.

ashum9:
Hi @siddharth7113 sir ,  I retried this from scratch while setting up the repo for the very first time in a clean directory.
I have attached a screenshot showing the full terminal reproduction. This is a rough first time developer experience because the very first import from a plain checkout crashes before the user even gets to `pip install -e .`.

image

Suggested fix is small: catch `importlib.metadata.PackageNotFoundError` in `pyaptamer/__init__.py` and fall back to something like `__version__ = "0+unknown"`.

siddharth7113:
Okay, but it is clearly stated in `README` you need to install it first, in fact you need to install any package first before using it, the suggested fix doesn't make sense.

See [here](https://github.com/gc-os-ai/pyaptamer?tab=readme-ov-file#development-install)

> Hi @siddharth7113 sir , I retried this from scratch while setting up the repo for the very first time in a clean directory. I have attached a screenshot showing the full terminal reproduction. This is a rough first time developer experience because the very first import from a plain checkout crashes before the user even gets to `pip install -e .`.
> 
> image
> Suggested fix is small: catch `importlib.metadata.PackageNotFoundError` in `pyaptamer/__init__.py` and fall back to something like `__version__ = "0+unknown"`.

PR #318, opened 2026-04-07
[BUG] Fix silent preprocessing dropping of short sequences and add defensive bounds

### LINKED ISSUE #284: [BUG] `rna2vec` dropping shorter sequences <3 without correct error

Rusheel86:
#### Reference Issues/PRs
Fixes #284.

This PR fixes a bug where extremely short biological sequences (length < 3) are silently skipped during tokenization and fed to deep neural network components as an empty tensor, triggering cryptic PyTorch C++ spatial exceptions (e.g. `Calculated padded input size per channel: (0 x 0)`). 

Changes implemented:
- Added a direct `ValueError` guard in _rna.py's `rna2vec()` function so unsupported sequence lengths clearly raise an understandable error instead of silently returning an empty numpy array.
- Added a centralized `validate_paired_tensors` structural guard rail at the entry point of `AptaTrans.forward_imap()` (inside _model.py) to ensure the neural networks inherently reject empty inputs or `NaN` features, regardless of pipeline or API origin.

- [x] **Validation Scope:** Confirm that `validate_tensor` effectively catches the edge cases (empty shapes and NaNs) without blocking valid input matrices.
- [x] **Placement:** This PR follows the explicit review guidance to catch the sequence violation natively in _rna.py while providing a strictly scoped safety net at the `AptaTrans` network boundary.

No but I manually verified these failure conditions with a local reproduction script to confirm that the sequence-length `ValueError` correctly pre-empts the pipeline from wrapping empty sequences. The Python code now handles this before hitting the C++ backend.

#### Any other comments?
This approach guarantees the pipeline fails-fast right at the data processing layer as suggested, significantly improving the developer experience for researchers using the pyaptamer API with fragmented biological sequences.

#### PR checklist

- [ ] Added/modified tests *(Relies on existing suite coverage for boundary testing).*


pyaptamer/utils/_rna.py
```diff
@@ -145,6 +145,11 @@ def rna2vec(
         if sequence_type == "rna":
             sequence = dna2rna(sequence)
 
+        if len(sequence) < 3:
```
siddharth7113: Instead of hardcoding the check here we can directly modify `converted` part , like this 

```python 
if not any(converted):  
    raise ValueError(
        f"Sequence '{sequence}' is too short to produce valid triplets. "
        f"Minimum length is 3."
    )

pyaptamer/utils/_validation.py
siddharth7113, line 1: Instead of creating a new file, make few line change just in`forward_imap` that should be sufficient.

siddharth7113 (changes requested):

siddharth7113:
Also tests are failing, please make correction in those too.

Rusheel86:
I’ve updated the PR  :

- Removed `_validation.py` and moved the logic directly into rna2vec (using the `if not any(converted)` check) and `forward_imap` as requested.

- I've had to update `test_aptamer.py` and `test_rna.py`. Previously, these tests were designed to expect silent failures (empty arrays or float outputs). I’ve updated them to use `pytest.raises(ValueError)` to align with the new validation logic.

- Adjusted the `MockAptaNetPipeline` in the test suite to mirror this new rejection behavior.

All tests are passing locally, and I've ensured the formatting matches the project's standards via Ruff. Tho I'm not sure if changing the tests this way is proper so pls let me know.

Rusheel86:
I've also modified `aptatrans_tutorial.ipynb`  a bit as only the .ipynb CI check was failing as it contains at least one sequence of length 2 ("GG").
Before our fix, `rna2vec` would just silently process "GG" into an empty list and create corrupted tensors. Now, doing its job correctly, our newly added defensive validation safely throws the `ValueError` when it hits that "GG" sequence  but this stops the tutorial notebook during the CI checks. Let me know if this was correct.

siddharth7113:
@NennoMP Would appreciate your feedback on this one !

NennoMP:
As far as I understood the issue occurs when we have aptamer sequences of `length < 3`. 

Not sure whether this is a problem. Do aptamer sequences with just 1 or 2 nucleotides (i.e., length 1 or 2) even exist? To the best of my knowledge, they do not. These checks won't do anything if the dataset is adequately curated and preprocessed. Indeed, we never ran into a crash when training AptaTrans on its own dataset.

@siddharth7113 What do you think? This issue would occur only if AptaTrans is given a badly curated dataset, containing sequences of `length < 3`, which, relastically, do not even represent real aptamers.

siddharth7113:
> As far as I understood the issue occurs when we have aptamer sequences of `length < 3`.
> 
> Not sure whether this is a problem. Do aptamer sequences with just 1 or 2 nucleotides (i.e., length 1 or 2) even exist? To the best of my knowledge, they do not. These checks won't do anything if the dataset is adequately curated and preprocessed. Indeed, we never ran into a crash when training AptaTrans on its own dataset.
> 
> @siddharth7113 What do you think? This issue would occur only if AptaTrans is given a badly curated dataset, containing sequences of `length < 3`, which, relastically, do not even represent real aptamers.

Yes, afaik that is the only case where it fails when `sequence < 3` , This is true it would only happen when the dataset is badly curated, but I am little bit conflicted, because if dataset is bad, Should be the one giving explicit warning to the user? Or the onus lies on the user itself? I feel these checks would atleast inform the end user about it, but again, I am not sure what is the possibility of happening in the real use case?

NennoMP:
> > As far as I understood the issue occurs when we have aptamer sequences of `length < 3`.
> > Not sure whether this is a problem. Do aptamer sequences with just 1 or 2 nucleotides (i.e., length 1 or 2) even exist? To the best of my knowledge, they do not. These checks won't do anything if the dataset is adequately curated and preprocessed. Indeed, we never ran into a crash when training AptaTrans on its own dataset.
> > @siddharth7113 What do you think? This issue would occur only if AptaTrans is given a badly curated dataset, containing sequences of `length < 3`, which, relastically, do not even represent real aptamers.
> 
> Yes, afaik that is the only case where it fails when `sequence < 3` , This is true it would only happen when the dataset is badly curated, but I am little bit conflicted, because if dataset is bad, Should be the one giving explicit warning to the user? Or the onus lies on the user itself? I feel these checks would atleast inform the end user about it, but again, I am not sure what is the possibility of happening in the real use case?

But at that point I feel like it should be responsibility of classes related to data preprocessing, data loading, and so on. It should not be the responsibility of each model, otherwise you end up with a lot of duplicated logic for checking data validity.

siddharth7113:
> > > As far as I understood the issue occurs when we have aptamer sequences of `length < 3`.
> > > Not sure whether this is a problem. Do aptamer sequences with just 1 or 2 nucleotides (i.e., length 1 or 2) even exist? To the best of my knowledge, they do not. These checks won't do anything if the dataset is adequately curated and preprocessed. Indeed, we never ran into a crash when training AptaTrans on its own dataset.
> > > @siddharth7113 What do you think? This issue would occur only if AptaTrans is given a badly curated dataset, containing sequences of `length < 3`, which, relastically, do not even represent real aptamers.
> > 
> > 
> > Yes, afaik that is the only case where it fails when `sequence < 3` , This is true it would only happen when the dataset is badly curated, but I am little bit conflicted, because if dataset is bad, Should be the one giving explicit warning to the user? Or the onus lies on the user itself? I feel these checks would atleast inform the end user about it, but again, I am not sure what is the possibility of happening in the real use case?
> 
> But at that point I feel like it should be responsibility of classes related to data preprocessing, data loading, and so on. It should've be the responsibility of each model, otherwise you end up with a lot of duplicated logic for checking data validity.

OK yeah makes sense, then this would be something we do after #441 , since there is still a lot of stuff to do to decouple the whole pipeline for `Aptatrans`

siddharth7113:
@Rusheel86 At this time I am closing the PR since this would require some other parts to be implemented, but thanks for flagging the issu.

PR #329, opened 2026-04-08
Make dna2rna conversion linear time

### LINKED ISSUE #323: [BUG] dna2rna() has O(n²) performance due to redundant str.replace() calls inside loop
**Describe the bug**

Code location
```python
for char in result:
    if char not in "ACGU":
return result
```

**To Reproduce**

```python
import time
    t0 = time.perf_counter()
    dna2rna(seq)
```
###  Benchmark Results

**Expected behavior**
Suggested fix
```python

```

**Additional context**

Python version: 3.12

Image

**Versions**
<details>

pyaptamer version: 0.1.0a1
Python version: 3.9.6
OS: macOS (darwin)

</details>

NandiniDhanrale:
Closes #323.

## Summary
- Replace repeated full-string replace calls in dna2rna with a single-pass conversion.
- Preserve existing T-to-U and unknown-to-N behavior.
- Add a regression test for repeated unknown nucleotides.

## Testing
- python -m pytest pyaptamer\utils\tests\test_rna.py
- python -m ruff check --no-cache pyaptamer\utils\_rna.py pyaptamer\utils\tests\test_rna.py

### DIFF
--- a/pyaptamer/utils/_rna.py
+++ b/pyaptamer/utils/_rna.py
@@ -32,10 +32,7 @@ def dna2rna(sequence: str) -> str:
     """
     # replace nucleotides 'T' with 'U'
     result = sequence.translate(str.maketrans("T", "U"))
-    for char in result:
-        if char not in "ACGU":
-            result = result.replace(char, "N")  # replace unknown nucleotides with 'N'
-    return result
+    return "".join(char if char in "ACGU" else "N" for char in result)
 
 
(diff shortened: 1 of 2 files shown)


siddharth7113:
Hi @NandiniDhanrale , Please follow contributors guidelines before opening PRs, and have discussion with maintainers in issues before opening PR, I am inclined to mark PRs as `ai-overuse`, I would recommend spending time reading and understanding codebase , figuring out issues and findings bugs, rather than opening multiple PRs that doesn't benifit anybody.

PR #330, opened 2026-04-08
Normalize RNA bases in AptaNet k-mer features

### LINKED ISSUE #322: [BUG] RNA k-mers with 'U' are silently ignored in generate_kmer_vecs
## Describe the bug

---

## To Reproduce
```python

# DNA baseline

# RNA sequence with U

```

```python
DNA_BASES = list("ACGT")
```

---

## Expected behavior

Possible expected behaviors:

---

## Additional context

---

## Versions
<details>

```python
import pyaptamer
print(pyaptamer.__version__)
# 0.1.0a1
```

</details>

---

NandiniDhanrale:
Closes #322.

## Summary
- Normalize RNA U bases to T before generating AptaNet k-mer features.
- Document that RNA-style input is mapped to the existing DNA k-mer convention.
- Add a regression test asserting AUG and ATG produce identical k-mer vectors.

## Testing
- python -m pytest pyaptamer\utils\tests\test_aptanet_utils.py
- python -m pytest pyaptamer\aptanet\tests\test_aptanet.py
- python -m ruff check --no-cache pyaptamer\utils\_aptanet_utils.py pyaptamer\utils\tests\test_aptanet_utils.py

### DIFF
--- a/pyaptamer/utils/_aptanet_utils.py
+++ b/pyaptamer/utils/_aptanet_utils.py
@@ -14,7 +14,8 @@ def generate_kmer_vecs(aptamer_sequence, k=4):
     Generate normalized k-mer frequency vectors for the aptamer sequence.
 
     For all possible k-mers from length 1 to k, count their occurrences in the sequence
-    and normalize to form a frequency vector.
+    and normalize to form a frequency vector. RNA bases are mapped to the AptaNet DNA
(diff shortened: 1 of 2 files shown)


siddharth7113:
This PR doesn't make sense, see the discussion [here](https://github.com/gc-os-ai/pyaptamer/issues/322#issuecomment-4204471755)

PR #331, opened 2026-04-08
Respect ignore_duplicates in MoleculeLoader

### LINKED ISSUE #312: [BUG] MoleculeLoader ignores ignore_duplicates when exporting sequences
**Describe the bug**
`MoleculeLoader` exposes an `ignore_duplicates` parameter, but `to_df_seq()` never uses it. As a result, duplicate amino-acid sequences are still returned even when callers explicitly request deduplication.

**To Reproduce**
```python
from pathlib import Path

from pyaptamer.data.loader import MoleculeLoader

root = Path('pyaptamer/datasets/data')
paths = [root / '1gnh.pdb', root / '1gnh.pdb']

loader = MoleculeLoader(paths, ignore_duplicates=True)
df = loader.to_df_seq()

print(df['sequence'].duplicated().any())
print(len(df), len(df['sequence'].drop_duplicates()))
```

**Expected behavior**
When `ignore_duplicates=True`, duplicate sequences should be removed while keeping the first occurrence, matching the documented parameter contract.

**Additional context**
This is a silent behavior mismatch in a public API: callers can pass the flag and receive duplicated rows without any warning. The implementation already documents the option and related utilities in the repo already support duplicate filtering, so the missing behavior is localized to `MoleculeLoader`.

**Versions**
<details>

```python
import pyaptamer
print(pyaptamer.__version__)
# 0.1.0a1
```

</details>

NandiniDhanrale:
Closes #312.

## Summary
- Apply ignore_duplicates in MoleculeLoader.to_df_seq().
- Remove duplicate sequences while keeping the first occurrence and preserving the existing MultiIndex output.
- Add tests for default duplicate preservation and requested deduplication.

## Testing
- python -m pytest pyaptamer\data\tests\test_loader.py
- python -m ruff check --no-cache pyaptamer\data\loader.py pyaptamer\data\tests\test_loader.py

### DIFF
--- a/pyaptamer/data/loader.py
+++ b/pyaptamer/data/loader.py
@@ -71,7 +71,11 @@ class MoleculeLoader:
 
         columns = ["sequence"] if self.columns is None else self.columns
 
-        return pd.DataFrame(sequences, index=index, columns=columns)
+        df = pd.DataFrame(sequences, index=index, columns=columns)
+        if self.ignore_duplicates:
+            df = df.drop_duplicates(subset=[columns[0]], keep="first")
+
+        return df
 
     def _determine_type(self, path):
         suffix = path.suffix.lower()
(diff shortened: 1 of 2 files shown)


siddharth7113:
Hi @NandiniDhanrale , This is already part of PR #259 , Please feel free to pick up another issue.

PR #336, opened 2026-04-08
[BUG] Fix augment_reverse() to compute actual reverse complement

### LINKED ISSUE #335: [ENH] augment_reverse() computes reverse but docstrings mentions `complement` which should be updated.
**Describe the bug**

`augment_reverse()` in `pyaptamer/utils/_augment.py` claims to augment sequences by adding their **reverse complement**, but the implementation only reverses the sequences without computing the Watson–Crick complement (A↔T/U, C↔G).

This means that when `APIDataset` (in `pyaptamer/datasets/dataclasses/_api.py`) uses `augment_reverse()` to augment aptamer training data, the augmented sequences are biologically meaningless reversals rather than reverse complements. Reverse complement is the standard data augmentation technique for nucleic acid sequences because the reverse complement strand is biologically related to the original via base-pairing rules.

**To Reproduce**

```python
import numpy as np
from pyaptamer.utils._augment import augment_reverse

sequences = np.array(["AACG"])
result = augment_reverse(sequences)
print(result[0])
# Current output: ['AACG' 'GCAA']  (just reversed)
# Expected output: ['AACG' 'CGUU']  (reverse complement, RNA convention)
```

**Expected behavior**

`augment_reverse()` should compute the reverse complement of each sequence:

1. Compute the Watson–Crick complement (A↔U/T, C↔G)
2. Reverse the complemented sequence

For example:
- `"AACG"` → complement `"TTGC"` → reverse `"CGTT"` (DNA) or equivalently `"CGUU"` after downstream T→U conversion

**Additional context**

- The docstring explicitly says *"reverse complement"* in both `augment_reverse()` and `APIDataset._prepare_data()`.
- The existing tests only verify reversal behavior because they were written to match the (incorrect) implementation.
- The function is used exclusively in `APIDataset._prepare_data()` for AptaTrans training data augmentation.

**Versions**
<details>

pyaptamer version: 0.1.0a1

</details>

officialasishkumar:
#### Reference Issues/PRs
Fixes #335.

`augment_reverse()` in `pyaptamer/utils/_augment.py` was documented as computing the reverse complement of nucleotide sequences, but the implementation only reversed the strings without applying Watson–Crick complement rules (A↔T/U, C↔G).

This fix adds a complement translation table and applies it before reversal, producing biologically correct reverse complements. The table uses DNA convention (A→T, T→A, U→A, C→G, G→C), which works for both DNA and RNA inputs since the downstream `dna2rna()` function handles T→U normalization. Non-nucleotide characters pass through unchanged.

The fix affects `APIDataset._prepare_data()` in `pyaptamer/datasets/dataclasses/_api.py`, which uses `augment_reverse()` to augment aptamer training data for AptaTrans. Previously, the augmented sequences were biologically meaningless reversals; now they are proper reverse complements.

- Whether the DNA-convention complement table (A→T instead of A→U) is the right choice given the downstream `dna2rna()` conversion
- Whether additional integration-level tests for `APIDataset` augmentation are needed

Yes, the test suite in `pyaptamer/utils/tests/test_augment.py` was updated with five focused tests:
- DNA reverse complement correctness
- RNA reverse complement correctness (verifying DNA-convention output)
- Multiple array handling
- Non-nucleotide character passthrough
- Empty array edge case

#### Any other comments?

Full test suite passes locally: 338 passed, 3 skipped.

#### PR checklist
- [x] Added/modified tests

### DIFF
--- a/pyaptamer/utils/_augment.py
+++ b/pyaptamer/utils/_augment.py
@@ -3,9 +3,22 @@ __all__ = ["augment_reverse"]
 
 import numpy as np
 
+# Watson-Crick complement table covering both DNA (T) and RNA (U) nucleotides.
+# Characters not in the table (e.g. 'N', non-biological characters) are left unchanged.
+_COMPLEMENT = str.maketrans("ACGTUacgtu", "TGCAAtgcaa")
+
+
+def _reverse_complement(seq: str) -> str:
+    """Return the reverse complement of a nucleotide sequence."""
+    return seq.translate(_COMPLEMENT)[::-1]
+
 
 def augment_reverse(*sequence_arrays: np.ndarray) -> tuple[np.ndarray, ...]:
     """Augment arrays of sequences by adding their reverse complement.
+
+    Computes the Watson-Crick reverse complement of each sequence using
+    standard base-pairing rules (A↔T/U, C↔G) and appends the results
+    to the original arrays.
 
     Parameters
     ----------
@@ -20,10 +33,10 @@ def augment_reverse(*sequence_arrays: np.ndarray) -> tuple[np.ndarray, ...]:
     """
     results = []
     for sequences in sequence_arrays:
-        # create array of reversed sequences
-        reversed_sequences = np.array([seq[::-1] for seq in sequences])
-        # concatenate original and reversed sequences
-        result = np.concatenate([sequences, reversed_sequences])
+        # compute the reverse complement of each sequence
+        rc_sequences = np.array([_reverse_complement(seq) for seq in sequences])
+        # concatenate original and reverse-complemented sequences
+        result = np.concatenate([sequences, rc_sequences])
         results.append(result)
 
     return tuple(results)
--- a/pyaptamer/utils/tests/test_augment.py
+++ b/pyaptamer/utils/tests/test_augment.py
@@ -7,33 +7,75 @@ import numpy as np
 from pyaptamer.utils._augment import augment_reverse
 
 
-def test_augment_reverse_single_array():
-    sequences = np.array(["AAC", "BBB", "ATCG"])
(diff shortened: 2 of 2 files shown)


NennoMP:
What does the Watson–Crick complement rule have to do with the data augmentation of `augment_reverse()`? The goal is to provide the model with both the original sequence and the reversed one, as in the original implementation (see `augment_seqset()` at [link](https://github.com/PNUMLB/AptaTrans/blob/59a8a8f618edde57a72555950b7c61c9c64b54fe/utils.py)) and as mentioned in the AptaTrans paper.

If reverse complement (which was meant in the string sense) causes confusion due to technicalities related to biology then we should fix the docstring, not the method.

Please refrain from opening random PRs for bugs that are non-existent. At least open an issue and discuss with maintainers.

siddharth7113:
@NennoMP I would say this was partly my fault, since I got confused as well , but yes @officialasishkumar please have discussion with maintainers before opening the PR, most algorithms are scientific implementations, and sometimes researchers make choices which might not be clearly communicated through code and overlooked, in this case , the original implementation is correct and what they intended, it would have been better if this was discussed before opening the PR.

ChandanKT-git:
> @NennoMP I would say this was partly my fault, since I got confused as well , but yes @officialasishkumar please have discussion with maintainers before opening the PR, most algorithms are scientific implementations, and sometimes researchers make choices which might not be clearly communicated through code and overlooked, in this case , the original implementation is correct and what they intended, it would have been better if this was discussed before opening the PR.

@NennoMP @siddharth7113 
I was confused as well. Shall I open an Issue and PR to fix the docstring from 'reverse complement' to simply 'reversed sequences' so future contributors don't get tripped up by the terminology?

PR #337, opened 2026-04-08
[BUG] Fix dna2rna() O(n²) performance and docstring parameter name

### LINKED ISSUE #323: [BUG] dna2rna() has O(n²) performance due to redundant str.replace() calls inside loop

officialasishkumar:
#### Reference Issues/PRs
Closes #323.

`dna2rna()` in `pyaptamer/utils/_rna.py` used `str.replace()` inside a `for char in result` loop to replace unknown nucleotides with `'N'`. This caused O(n²) performance because:

1. `str.replace()` scans the entire string for each unique unknown character
2. The loop iterates over the original string snapshot while `result` is reassigned on each replacement

The fix replaces the two-step approach (first `str.translate` for T→U, then loop for unknowns) with a single-pass generator expression that handles both T→U conversion and unknown→N replacement in one character-by-character scan, achieving O(n) time complexity.

Benchmark comparison (from issue #323):
| Input Size | Before (loop + `.replace()`) | After (`join` + genexpr) | Speedup |
| 100,000    | ~0.30s                       | ~0.006s                  | ~50x    |
| 500,000    | ~7.40s                       | ~0.028s                  | ~264x   |

Also fixes the docstring parameter name from `seq` to `sequence` to match the actual function signature.

- The single-pass approach preserves exact behavior (T→U, unknown→N, valid characters unchanged)
- All 20 existing RNA tests pass plus 1 new regression test for repeated unknowns

Yes, added `test_dna2rna_repeated_unknowns` which verifies:
- All-unknown sequences are correctly replaced with 'N'
- Mixed valid/unknown sequences are handled correctly
- A 10,000-character unknown sequence runs without timeout (would take ~0.3s with the old O(n²) code)

#### Any other comments?

Full test suite: 338 passed, 3 skipped. All doctests pass.

#### PR checklist
- [x] Added/modified tests

### DIFF
--- a/pyaptamer/utils/_rna.py
+++ b/pyaptamer/utils/_rna.py
@@ -30,12 +30,9 @@ def dna2rna(sequence: str) -> str:
     str
         The converted RNA sequence.
     """
-    # replace nucleotides 'T' with 'U'
-    result = sequence.translate(str.maketrans("T", "U"))
-    for char in result:
-        if char not in "ACGU":
-            result = result.replace(char, "N")  # replace unknown nucleotides with 'N'
-    return result
+    _VALID = frozenset("ACGU")
+    # replace 'T' with 'U' and any unknown nucleotides with 'N' in a single pass
+    return "".join("U" if c == "T" else (c if c in _VALID else "N") for c in sequence)
 
 
 def generate_nplets(letters: list[str], repeat: int | Iterable[int]) -> dict[str, int]:
--- a/pyaptamer/utils/_rna.py
+++ b/pyaptamer/utils/_rna.py
@@ -11,6 +11,8 @@ from itertools import product
 
 import numpy as np
 
+_VALID_NUCLEOTIDES = frozenset("ACGU")
+
 
 def dna2rna(sequence: str) -> str:
     """
@@ -30,9 +32,10 @@ def dna2rna(sequence: str) -> str:
     str
         The converted RNA sequence.
     """
-    _VALID = frozenset("ACGU")
-    # replace 'T' with 'U' and any unknown nucleotides with 'N' in a single pass
-    return "".join("U" if c == "T" else (c if c in _VALID else "N") for c in sequence)
+    # replace nucleotides 'T' with 'U'
+    result = sequence.translate(str.maketrans("T", "U"))
+    result = "".join(char if char in _VALID_NUCLEOTIDES else "N" for char in result)
+    return result
 
 
 def generate_nplets(letters: list[str], repeat: int | Iterable[int]) -> dict[str, int]:
--- a/pyaptamer/utils/tests/test_rna.py
+++ b/pyaptamer/utils/tests/test_rna.py
@@ -36,6 +36,18 @@ def test_dna2rna_edge_cases():
     assert dna2rna("AcGt") == "ANGN"
 
 
+def test_dna2rna_repeated_unknowns():
+    """Check that repeated unknown characters are all replaced with 'N'."""
+    # all unknown characters should each become 'N'
+    assert dna2rna("XYZXYZ") == "NNNNNN"
+    # mix of valid and repeated unknowns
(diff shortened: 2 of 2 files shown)


pyaptamer/utils/tests/test_rna.py
```diff
@@ -36,6 +36,18 @@ def test_dna2rna_edge_cases():
...
+def test_dna2rna_repeated_unknowns():
+    """Check that repeated unknown characters are all replaced with 'N'."""
+    # all unknown characters should each become 'N'
+    assert dna2rna("XYZXYZ") == "NNNNNN"
+    # mix of valid and repeated unknowns
+    assert dna2rna("AXXGTT") == "ANNGUU"
+    # large sequence with many unknowns runs in linear time
+    large_unknown = "X" * 10_000
+    result = dna2rna(large_unknown)
+    assert result == "N" * 10_000
```
NennoMP: I believe these tests are not needed. There are already covered by the following:

```python
@pytest.mark.parametrize(
    "dna, expected_rna",
    [
        ("AAA", "AAA"),
        ("ACG", "ACG"),
        ("AAT", "AAU"),
        ("TTT", "UUU"),
        ("AAX", "AAN"),
        ("XXX", "NNN"),
    ],
)
def test_dna2rna(dna, expected_rna):
    """Check conversion of DNA to RNA nucleotides."""
    assert dna2rna(dna) == expected_rna
```

pyaptamer/utils/_rna.py
```diff
@@ -22,20 +22,17 @@ def dna2rna(sequence: str) -> str:
 
     Parameters
     ----------
-    seq : str
+    sequence : str
```
NennoMP, line 27: Please do not change variable names. Revert this change.

NennoMP (changes requested):

NennoMP:
@officialasishkumar Minimal changes needed. Address all feedback and I will merge.

pyaptamer/utils/_rna.py
```diff
@@ -22,20 +22,17 @@ def dna2rna(sequence: str) -> str:
...
     str
         The converted RNA sequence.
     """
-    # replace nucleotides 'T' with 'U'
-    result = sequence.translate(str.maketrans("T", "U"))
-    for char in result:
-        if char not in "ACGU":
-            result = result.replace(char, "N")  # replace unknown nucleotides with 'N'
-    return result
+    _VALID = frozenset("ACGU")
```
NennoMP: Why not just call this variable `valid` (even better `valid_nucleotides`)? Uppercase is more standard for global variables within the module, and preprending `_` is standard for private variables/methods.

pyaptamer/utils/tests/test_rna.py
officialasishkumar: Removed the redundant `test_dna2rna_repeated_unknowns` block.

pyaptamer/utils/_rna.py
officialasishkumar: Rebased onto current `main`; this PR no longer changes the docstring parameter name.

pyaptamer/utils/_rna.py
officialasishkumar: Moved this to module-private `_VALID_NUCLEOTIDES` and kept the existing `result`/`char` naming in the implementation.

officialasishkumar:
Addressed the requested feedback in `37ee05f` after rebasing onto current `main`.

NennoMP (approved):

PR #341, opened 2026-04-08
fix: integrate pdb_to_aaseq fallback in MoleculeLoader

WHOIM1205:
### Summary
This PR fixes an issue where `MoleculeLoader` fails to extract sequences from valid PDB files that do not contain `SEQRES` headers. In such cases, the loader returned an empty DataFrame, leading to downstream failures or silent data issues.

---

### What was happening
- `MoleculeLoader._load_pdb_seq` relied only on:
  `SeqIO.parse(handle, "pdb-seqres")`
- If `SEQRES` records were missing (common in many real-world PDBs), no sequences were extracted.
- This resulted in:
  - Empty datasets (`0` sequences)
  - Crashes during training (`Found array with 0 sample(s)`)
  - Silent data issues in pipelines

---

### What’s fixed
- Replaced direct `SeqIO` parsing with `pdb_to_aaseq`, which:
  - Uses `SEQRES` when available
  - Falls back to structure-based parsing (via PPBuilder) when missing
- Standardized output column:
  - `chain` → `chain_id`

---

### Code Changes
```python
from pyaptamer.utils._pdb_to_aaseq import pdb_to_aaseq

df = pdb_to_aaseq(path, return_type="pd.df", ignore_duplicates=self.ignore_duplicates)
return df.rename(columns={"chain": "chain_id"})
```

---

### Impact
- Fixes failure on PDB files without `SEQRES`
- Prevents empty outputs and downstream crashes
- Aligns `MoleculeLoader` with existing utility behavior
- Improves robustness for real-world datasets

---

### How to verify
```python
from pathlib import Path
from pyaptamer.data.loader import MoleculeLoader

mol = MoleculeLoader(Path("pyaptamer/datasets/data/1gnh_no_seqres.pdb"))
df = mol.to_df_seq()

print(len(df))  # Before: 0 | After: 10
```

---

### Notes
- No breaking API changes
- Existing behavior for PDBs with `SEQRES` remains unchanged

### DIFF
--- a/pyaptamer/data/loader.py
+++ b/pyaptamer/data/loader.py
@@ -98,14 +98,7 @@ class MoleculeLoader:
         pandas.DataFrame
             DataFrame with columns ``["chain_id", "sequence"]``.
         """
-        with open(path) as handle:
-            seqres_records = list(SeqIO.parse(handle, "pdb-seqres"))
+        from pyaptamer.utils._pdb_to_aaseq import pdb_to_aaseq
 
-        records = [
-            {
-                "chain_id": record.id.split(":")[1] if ":" in record.id else record.id,
-                "sequence": str(record.seq),
-            }
-            for record in seqres_records
-        ]
-        return pd.DataFrame.from_records(records, columns=["chain_id", "sequence"])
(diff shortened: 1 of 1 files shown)


satvshr:
Unnecessary change

PR #343, opened 2026-04-08
fix(utils): generate reverse complement instead of reverse

WHOIM1205:
### Description
This PR fixes a critical issue in the `augment_reverse` function where nucleotide sequences were being reversed but not complemented.

Previously, the implementation used Python string slicing (`[::-1]`), which only reversed sequences (e.g., `ATCG → GCTA`). However, the intended behavior—as stated in the docstring—is to generate reverse complements (e.g., `ATCG → CGAT`).

This resulted in biologically invalid augmented data and silent data corruption.

---

### What was fixed
- Added a nucleotide translation step using:
  str.maketrans("ACGTUacgtu", "TGCAAtgcaa")

- Updated sequence transformation to:
  seq.translate(trans)[::-1]

- Kept existing numpy structure and concatenation logic unchanged

---

### Why this matters
- Prevents silent data corruption in datasets
- Ensures biologically valid reverse complement sequences
- Improves downstream ML model reliability
- Eliminates incorrect motif learning

---

### Changes made
- pyaptamer/utils/_augment.py
  - Fixed reverse logic to include complement step

- pyaptamer/utils/tests/test_augment.py
  - Updated expected outputs

---

### Test Verification
- Updated unit tests to reflect correct behavior
- All tests passing:
  pytest pyaptamer/utils/tests/test_augment.py

---

### Example
augment_reverse(np.array(["ATCG"]))

Before:
["ATCG", "GCTA"]

After:
["ATCG", "CGAT"]

---

### Summary
Fixes a critical biological bug by ensuring sequence augmentation produces true reverse complements instead of simple reversed strings.

### DIFF
--- a/pyaptamer/utils/_augment.py
+++ b/pyaptamer/utils/_augment.py
@@ -18,10 +18,11 @@ def augment_reverse(*sequence_arrays: np.ndarray) -> tuple[np.ndarray, ...]:
         A tuple of arrays, each containing sequences with their reverse complements
         added.
     """
+    trans = str.maketrans("ACGTUacgtu", "TGCAAtgcaa")
     results = []
     for sequences in sequence_arrays:
         # create array of reversed sequences
-        reversed_sequences = np.array([seq[::-1] for seq in sequences])
+        reversed_sequences = np.array([seq.translate(trans)[::-1] for seq in sequences])
         # concatenate original and reversed sequences
         result = np.concatenate([sequences, reversed_sequences])
         results.append(result)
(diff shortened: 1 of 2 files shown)


siddharth7113:
Hi, Is there any issue open about this? What you provide as example is also wrong. The intended behaviour should have been `A<->T` `C<->G`

>Example
augment_reverse(np.array(["ATCG"]))
>Before:
["ATCG", "GCTA"]
>After:
["ATCG", "CGAT"]

siddharth7113:
Okay clsoing as this is duplicate of #336

PR #347, opened 2026-04-09
[BUG] Raise error for non-DNA characters in generate_kmer_vecs (#322)

### LINKED ISSUE #322: [BUG] RNA k-mers with 'U' are silently ignored in generate_kmer_vecs

codeit-ronit:
## Title:
  [BUG] Raise error for non-DNA characters in generate_kmer_vecs 

  Fixes #322

## Description

`generate_kmer_vecs` silently ignores RNA k-mers containing 'U' because , it only recognizes DNA bases (A, C, G, T). Instead of silently producing zero frequencies, the function now raises a `ValueError` when the sequence contains non-DNA characters.

As discussed in #322, this function is DNA-only by design — the fix raises an error rather than normalizing RNA input.

## Changes

  - Added input validation in `generate_kmer_vecs` to reject non-DNA characters
  - Added `Raises` section to the docstring
  - Added test coverage for `generate_kmer_vecs` (previously had none)

  ## Tests

  - 11 new tests in `test_aptanet_utils.py`
  - All existing tests pass with no regressions

### DIFF
--- a/pyaptamer/utils/_aptanet_utils.py
+++ b/pyaptamer/utils/_aptanet_utils.py
@@ -28,8 +28,20 @@ def generate_kmer_vecs(aptamer_sequence, k=4):
     np.ndarray
         1D numpy array of normalized frequency vector for all possible k-mers from
         length 1 to k.
+
+    Raises
+    ------
+    ValueError
+        If the sequence contains non-DNA characters (e.g., 'U' from RNA).
     """
     DNA_BASES = list("ACGT")
+
+    invalid_chars = set(aptamer_sequence.upper()) - set(DNA_BASES)
+    if invalid_chars:
+        raise ValueError(
+            f"Sequence contains non-DNA characters: {invalid_chars}. "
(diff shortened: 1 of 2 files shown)


satvshr:
Parking this because the person who made the issue wanted to work on this first.

prashu0705:
Thanks for working on this. Since I originally opened #322 and the maintainer has parked this PR for me, I’m going to open a focused PR for the issue.

I’ll review your approach first so I can avoid missing any important edge cases. Appreciate the work you already put into this.

satvshr:
@prashu0705 please answer the question ive left on the issue along with what changes youll be making, thanks

Issue #302, opened 2026-04-05
[BUG] `MaskedDataset.__getitem__` initializes `y_masked` from `x` instead of `y`


officialasishkumar:
**Describe the bug**

In `pyaptamer/datasets/dataclasses/_masked.py`, line 154 of `__getitem__` initializes `y_masked` by cloning `x` instead of `y`:

```python
x = torch.tensor(self.x[index], dtype=torch.int64)
y = torch.tensor(self.y[index], dtype=torch.int64)

x_masked = x.clone().detach()
y_masked = x.clone().detach()  # BUG: should be y.clone().detach()
```

This means the target tensor `y_masked` is a copy of the input sequence rather than the actual target sequence. The non-masked positions are then zeroed out from the wrong base tensor.

**To Reproduce**

```python
from pyaptamer.datasets.dataclasses import MaskedDataset

# Use different x and y to make the bug visible
x = [[1, 2, 3, 4, 5, 0]]
y = [[10, 20, 30, 40, 50, 0]]  # different from x

dataset = MaskedDataset(x, y, max_len=6, mask_idx=99, masked_rate=0.5)
x_masked, y_masked, x_orig, y_orig = dataset[0]

# y_masked is derived from x, not y — the values 10, 20, 30, 40, 50 never appear in y_masked
```

**Expected behavior**

`y_masked` should be initialized from `y`, not `x`:
```python
y_masked = y.clone().detach()
```

**Additional context**

In typical masked language modeling (MLM) setups, `x` and `y` are often identical (the same sequence), so this bug may not have been caught in practice. However, the API explicitly accepts separate `x` and `y` arrays, and the variable is named `y_masked`, so it should clone `y`.

Additionally, `MaskedDataset` currently has no dedicated unit tests for `__getitem__`, which is why this went undetected.

**Versions**
<details>
0.1.0a1
</details>

siddharth7113:
cc @NennoMP

NennoMP:
@officialasishkumar Actually, the `__getitem__` method logic is the way it is because the original repository had `y_masked` cloned from `x`. See `Masked_Dataset` class in [https://github.com/PNUMLB/AptaTrans/blob/master/utils.py](https://github.com/PNUMLB/AptaTrans/blob/master/utils.py).

When AptaTrans was implemented we simply tried to replicate and refactor the authors' code. Hard to tell whether this is the intended behaviour or they did a mistake in their codebase.

siddharth7113:
[...]

PR #356, opened 2026-04-10
[BUG] Fix MaskedDataset y_masked initialized from x instead of y

kunal14901:
#### Reference Issues/PRs
Fixes a previously unreported bug in `MaskedDataset.__getitem__()`.

`MaskedDataset.__getitem__()` in `pyaptamer/datasets/dataclasses/_masked.py` initializes `y_masked` by cloning `x` instead of `y`:

Before (line 156):
```python
x_masked = x.clone().detach()
y_masked = x.clone().detach()  # bug: should be y
```

After:
```python
x_masked = x.clone().detach()
y_masked = y.clone().detach()
```

When `x != y`, the masked target returned for loss computation contains values from the input sequence instead of the actual target sequence. The return docstring says it returns "target sequence with non-masked positions set to 0", but it was returning the input sequence with non-masked positions zeroed.

In typical MLM usage `x == y` so the bug is masked, but the class API explicitly accepts different `x` and `y` arrays via its constructor.

- Whether the fix correctly initializes `y_masked` from `y`
- Whether the regression test adequately covers the divergent x/y case

Yes. Added `pyaptamer/datasets/tests/test_masked_dataset.py` with 4 tests:
- `test_masked_dataset_length` — basic `__len__` check
- `test_masked_dataset_getitem_returns_four_tensors` — output shape
- `test_masked_dataset_mismatched_lengths` — `ValueError` on mismatched x/y
- `test_y_masked_initialized_from_y_not_x` — regression test for this specific bug

#### Any other comments?
Local validation:
- `pytest pyaptamer/datasets/tests/test_masked_dataset.py` — 4 passed
- `pre-commit run --files pyaptamer/datasets/dataclasses/_masked.py pyaptamer/datasets/tests/test_masked_dataset.py` — all passed

#### PR checklist
- [x] Added/modified tests
```

### DIFF
--- a/pyaptamer/datasets/dataclasses/_masked.py
+++ b/pyaptamer/datasets/dataclasses/_masked.py
@@ -151,7 +151,7 @@ class MaskedDataset(Dataset):
         y = torch.tensor(self.y[index], dtype=torch.int64)
 
         x_masked = x.clone().detach()
-        y_masked = x.clone().detach()
+        y_masked = y.clone().detach()
 
         # non-padding positions (0 is padding)
         seq_len = torch.sum(x_masked > 0)
--- a/pyaptamer/datasets/tests/test_masked_dataset.py
+++ b/pyaptamer/datasets/tests/test_masked_dataset.py
@@ -1,0 +1,48 @@
+"""Tests for MaskedDataset."""
+
+import pytest
+
+from pyaptamer.datasets.dataclasses import MaskedDataset
+
+
+@pytest.fixture
+def basic_dataset():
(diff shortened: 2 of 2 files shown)


NennoMP:
The issue you are referring (#302) is not even open. It has been closed since cloning `y_masked` from `x` is expected behaviour as also observed in the authors' original code.

Closing this.

kunal14901:
Thanks for the context, wasn't aware #302 was already closed as expected behaviour. Will check existing issues before opening PRs going forward.

PR #359, opened 2026-04-10
[BUG] Replace assert with ValueError in MCTS _reconstruct

kunal14901:
#### Reference Issues/PRs
Fixes a previously unreported bug in `MCTS._reconstruct()`.

`_reconstruct()` in `pyaptamer/mcts/_algorithm.py` uses an `assert` to validate that the encoded sequence has even length. `assert` statements are stripped when running with `python -O`, silently removing this validation.

Replaced `assert` with a proper `ValueError` raise.

- Whether `ValueError` is the right exception type here

Yes. Added `test_reconstruct_odd_length_raises` to `pyaptamer/mcts/tests/test_mcts.py`.

#### Any other comments?
- `pytest pyaptamer/mcts/tests/test_mcts.py` — all passed
- `pre-commit run --all-files` — all passed

#### PR checklist
- [x] Added/modified tests

### DIFF
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -124,9 +124,10 @@ class MCTS(BaseObject):
         # if the sequence is not reconstructed yet, it should have an even length
         # because it should consist of pairs such as 'A_' and '_A' (i.e., nucleotide +
         # direction marker).
-        assert len(sequence) % 2 == 0, (
-            f"Encoded sequence must have even length, got {len(sequence)}."
-        )
+        if len(sequence) % 2 != 0:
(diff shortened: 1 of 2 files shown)


NennoMP:
We are intentionally using assert here because the even-length constraint is an internal check/condition, not something depending on user input. Essentially, odd length is something that should never happen unless there is a bug in the algorithm.

This change is not needed. Please, in the future, open an issue first and engage in a discussion with maintainers. Do not open a PR right away.

kunal14901:
Makes sense,  assert is the right tool for internal invariants. Noted on opening issues first, will do that going forward.

PR #361, opened 2026-04-10
[BUG]: Validate `n_iterations` in MCTS

### LINKED ISSUE #358: [BUG]: No Validation of `n_iterations` in MCTS
**Describe the bug**
`remember.md` explicitly documents that a `ValueError` was added to `MCTS.__init__` to raise when `n_iterations <= 0`. No such check exists in the code. Passing `n_iterations=0` causes the inner `for _ in range(0)` loop to never execute, but the outer `while` loop stalls indefinitely because `self.base` never progresses past an empty string.

**To Reproduce**
```python
def __init__(
    self,
    states: list[str] | None = None,
    depth: int = 20,
    n_iterations: int = 1000,
    experiment=None,
) -> None:
    self.experiment = experiment
    self.depth = depth
    self.n_iterations = n_iterations
    # No validation of n_iterations!
```
**Expected behavior:**
Silent infinite loop with `n_iterations=0` or negative values.

**Versions**
<details>
0.1.0a1
</details>

direkkakkar319-ops:
#### Reference Issues/PRs
Fixes #358

1. Expanded constructor input validation in `MCTS.__init__`:
* Raises `ValueError` when:- **
a. `n_iterations < 1`.
b. `depth < 1`.
c. `states` is empty.
d. `states` contains duplicates.

4. Added focused regression tests in `test_mcts.py`:
* Verifies invalid:-
a. `n_iterations` values (`0`, `-1`) are rejected.
b. `depth` values (`0`, `-1`) are rejected.
c.  `states` inputs (empty and duplicate entries) are rejected.

```python
def __init__(self, ...) -> None:
    if depth < 1:
        raise ValueError(f"`depth` must be >= 1, got {depth}.")
    if n_iterations < 1:
        raise ValueError(f"`n_iterations` must be >= 1, got {n_iterations}.")

    if states is None:
        states = ["A_", "C_", "G_", "U_", "_A", "_C", "_G", "_U"]
    elif not states:
        raise ValueError("`states` must contain at least one entry.")
    elif len(states) != len(set(states)):
        raise ValueError("`states` must contain unique entries.")
```

- [ ] Please focus on whether validating `depth`, `n_iterations`, and `states` in `MCTS.__init__` is the right and consistent fix for preventing invalid runtime behavior (including the infinite-loop path in `run()` for `n_iterations <= 0`).
- [ ] Please check whether `ValueError` is the correct API behavior for these invalid constructor inputs.
- [ ] Please check that the added regression tests are focused and cover the relevant invalid cases without over-scoping.

yes

```python
@pytest.mark.parametrize("depth", [0, -1])
def test_init_invalid_depth(self, depth):
    ...

@pytest.mark.parametrize("n_iterations", [0, -1])
def test_init_invalid_iterations(self, n_iterations):
    ...

def test_init_empty_states(self):
    ...

def test_init_duplicate_states(self):
    ...
```

#### PR checklist
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] 
- [x] Added/modified tests

py-test
image

notebooks diff
Screenshot 2026-04-11 004844

pre-commit
image

run-jupyter-notebooks
Image

### DIFF
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -74,6 +74,9 @@ class MCTS(BaseObject):
         n_iterations: int = 1000,
         experiment=None,
     ) -> None:
+        if n_iterations < 1:
+            raise ValueError(f"`n_iterations` must be >= 1, got {n_iterations}.")
+
         self.experiment = experiment
         self.depth = depth
         self.n_iterations = n_iterations
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -24,11 +24,13 @@ class MCTS(BaseObject):
     ----------
     states : list[str], optional, default=None
         Possible values for the nodes. Underscores indicate whether the values are
-        supposed to be prepended or appended to the sequence.
+        supposed to be prepended or appended to the sequence. Must be non-empty and
+        contain unique entries.
     depth : int, optional, default=20
-        Maximum depth of the search tree, also the length of the generated sequences.
+        Maximum depth of the search tree, also the length of the generated
+        sequences. Must be >= 1.
     n_iterations : int, optional, default=1000
-        Number of iterations per round for the MCTS algorithm.
+        Number of iterations per round for the MCTS algorithm. Must be >= 1.
     experiment : BaseAptamerEval, optional, default=None
         An instance of an experiment class definingthe goal function for the algorithm.
 
@@ -74,17 +76,23 @@ class MCTS(BaseObject):
         n_iterations: int = 1000,
         experiment=None,
     ) -> None:
+        if depth < 1:
+            raise ValueError(f"`depth` must be >= 1, got {depth}.")
         if n_iterations < 1:
             raise ValueError(f"`n_iterations` must be >= 1, got {n_iterations}.")
+
+        if states is None:
+            states = ["A_", "C_", "G_", "U_", "_A", "_C", "_G", "_U"]
+        elif not states:
+            raise ValueError("`states` must contain at least one entry.")
+        elif len(states) != len(set(states)):
+            raise ValueError("`states` must contain unique entries.")
 
         self.experiment = experiment
         self.depth = depth
         self.n_iterations = n_iterations
 
         super().__init__()
-
-        if states is None:
-            states = ["A_", "C_", "G_", "U_", "_A", "_C", "_G", "_U"]
         self.states = states
 
         self.root = TreeNode(
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -34,15 +34,6 @@ class MCTS(BaseObject):
     experiment : BaseAptamerEval, optional, default=None
         An instance of an experiment class definingthe goal function for the algorithm.
 
-    Raises
-    ------
-    ValueError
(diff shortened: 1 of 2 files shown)


direkkakkar319-ops:
Hi, @NennoMP could you please confirm these two review points and tick the boxes if they look good to you

direkkakkar319-ops:
Once they are checked, I’ll go ahead and open the PR

NennoMP:
I would say it's quite trivial for the user (after reading the docstrings) that there is no point in passing `n_iterations = 0`, same thing for `depth = 0`.

Anyway, if you plan to introduce constructor parameters' validation, what's the point of doing so only for `n_iterations` and completely ignore other parameters (e.g., `depth`). I would suggest to update both issue and PR and add input validation for all sensible parameters, not just one.

Additionally, refrain from tagging maintainers and asking for a reviewer when the PR has been open for just 1 hour. Give it a least 1 week before reaching out.

direkkakkar319-ops:
Thanks for the feedback, that makes sense. I agre that if constructor validation is introducd, it should be applied consistently to all sensible parameters rather than only `n_iterations ,` so I wiilll updat both the issue and the PR scope accordingly, including validation for cases such as `depth <= 0` .

direkkakkar319-ops:
Also noted on the review etiquette. 
I apologize for reaching out too early and for tagging maintainers after the PR had only been open for a short time.

direkkakkar319-ops:
I will avoid doing that again and will give futur PRs' more time before following up : )

pyaptamer/mcts/_algorithm.py
```diff
@@ -74,14 +76,23 @@ def __init__(
...
+        if n_iterations < 1:
+            raise ValueError(f"`n_iterations` must be >= 1, got {n_iterations}.")
+
+        if states is None:
+            states = ["A_", "C_", "G_", "U_", "_A", "_C", "_G", "_U"]
+        elif not states:
+            raise ValueError("`states` must contain at least one entry.")
+        elif len(states) != len(set(states)):
+            raise ValueError("`states` must contain unique entries.")
+
```
NennoMP, line 98: I would merge the checks for `None` and emtpy list and default to `states = ["A_", "C_", "G_", "U_", "_A", "_C", "_G", "_U"]` for both.

pyaptamer/mcts/_algorithm.py
```diff
@@ -74,14 +76,23 @@ def __init__(
         n_iterations: int = 1000,
         experiment=None,
     ) -> None:
```
NennoMP, line 78: Please add a docstrings for the raised exceptions in `numpydoc` style (https://numpydoc.readthedocs.io/en/latest/format.html).

NennoMP (changes requested):

pyaptamer/mcts/_algorithm.py
NennoMP: @direkkakkar319-ops I should've been more specific. I meant to add them in `__init__`. I believe in other classes in the codebase, such as AptaTransPipeline, we usually put raised exceptions in the initialization method rather then in the class docstring.

direkkakkar319-ops: Apologies for the misunderstanding, moved the Raiss section into `__init__` to match the `AptaTransPipeline` convention  .

NennoMP (approved):

PR #362, opened 2026-04-10
[BUG] Replace AlphaDropout with Dropout in AptaNet layers (ReLU misma…

kunal14901:
#### Reference Issues/PRs
Fixes a previously unreported bug in `aptanet_layer()`.

`aptanet_layer()` in `pyaptamer/aptanet/_aptanet_nn.py` pairs `nn.AlphaDropout` with `nn.ReLU`. Per the PyTorch docs, `AlphaDropout` is designed exclusively for `SELU` activations to maintain the self-normalizing property. With `ReLU`, its noise injection produces wrong output statistics, silently degrading regularization across all 8 hidden layers in `AptaNetMLP`.

Changed `nn.AlphaDropout` to `nn.Dropout` to match the `ReLU` activation.

- Whether `Dropout + ReLU` is preferred over `AlphaDropout + SELU`

Yes. Added `test_aptanet_layer_uses_standard_dropout` to `pyaptamer/aptanet/tests/test_aptanet.py`.

#### Any other comments?
- `pytest pyaptamer/aptanet/tests/test_aptanet.py` — passed
- `pre-commit run --all-files` — passed

#### PR checklist
- [x] Added/modified tests

### DIFF
--- a/pyaptamer/aptanet/_aptanet_nn.py
+++ b/pyaptamer/aptanet/_aptanet_nn.py
@@ -7,7 +7,7 @@ import torch.nn as nn
 def aptanet_layer(input_dim, output_dim, dropout, lazy=False):
     """
     Create a single AptaNet layer composed of a linear transformation,
-    ReLU activation, and AlphaDropout.
+    ReLU activation, and Dropout.
 
     Parameters
     ----------
@@ -18,7 +18,7 @@ def aptanet_layer(input_dim, output_dim, dropout, lazy=False):
         Size of each output sample (i.e., number of neurons in the layer).
 
     dropout : float
-        Dropout probability for AlphaDropout. Must be between 0 and 1.
+        Dropout probability for Dropout. Must be between 0 and 1.
 
     lazy : bool, optional
         If True, use `nn.LazyLinear` instead of `nn.Linear`, allowing the input
@@ -30,13 +30,13 @@ def aptanet_layer(input_dim, output_dim, dropout, lazy=False):
         A sequential container with:
         - Linear or LazyLinear layer
         - ReLU activation
-        - AlphaDropout layer
+        - Dropout layer
     """
     linear = nn.LazyLinear(output_dim) if lazy else nn.Linear(input_dim, output_dim)
     return nn.Sequential(
         linear,
         nn.ReLU(),
-        nn.AlphaDropout(dropout),
+        nn.Dropout(dropout),
     )
 
 
--- a/pyaptamer/aptanet/_aptanet_nn.py
+++ b/pyaptamer/aptanet/_aptanet_nn.py
@@ -9,6 +9,10 @@ def aptanet_layer(input_dim, output_dim, dropout, lazy=False):
     Create a single AptaNet layer composed of a linear transformation,
     ReLU activation, and Dropout.
 
+    Note: This implementation uses standard Dropout (not AlphaDropout) with ReLU.
+    AlphaDropout is specifically designed for SELU activations to maintain
+    self-normalizing properties; with ReLU, it produces incorrect output statistics.
+
     Parameters
     ----------
     input_dim : int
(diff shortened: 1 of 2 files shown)


NennoMP:
FYI @satvshr.

The original AptaNet implementation had `AlphaDropout` along with `ReLU` activation. However, you are right and `AlphaDropout` is specifically designed for `SELU` activations, as per [R1].

We could do one of three things:
1. Keep it as is, since it's the original implementation, although fundamentally wrong.
2. We can replace `AlphaDropout` with `Dropout`, which you did.
3. We can replace `ReLU` with `SELU` activations.

Not sure what would be best between (2) and (3), I need to think about it.

---
[R1] Klambauer et al., _Self-Normalizing Neural Networks_, NeurIPS (2017).

satvshr:
I would keep it as it is @NennoMP, if it does not cause failures and still "works", it is okay? Though I'd like to know if you have a strong opinion regarding it.

siddharth7113:
> ****
> 
> * Replace `nn.AlphaDropout` with `nn.Dropout` in `aptanet_layer()` while keeping `ReLU`.

@NennoMP What's the difference between the 2? And how does it affect NN?
@satvshr  I would merge this if the intended docs, say that the correct way to do it is `AlphaDropout` with a note about this change in the code.

satvshr:
> @satvshr I would merge this if the intended docs, say that the correct way to do it is `AlphaDropout` with a note about this change in the code.

Hmm I don't agree, that would be changing the architecture no matter how minor the change is, don't really see a problem if it is working.

NennoMP:
> @NennoMP What's the difference between the 2? And how does it affect NN? @satvshr I would merge this if the intended docs, say that the correct way to do it is `AlphaDropout` with a note about this change in the code.

In (3) you would have the self-normalizing property from [R1], which you can have only by combining AlphaDropout with SELU activations. In (1) and (2) you don't have this property.

Overall I lean towards (2), but open to discussion. The original version is fundamentally wrong, AlphaDropout is specifically designed with SELU activations in mind. While (3) would also be a solution, it would introduce self-normalization which wasn't in the original implementation. 

In general, even if the model works, in some tasks training could be unstable or performance underwhelming. Also, conceptually, the model should work better this way. Thus, I don't see issues for reproducing the paper results. I definitely suggest adding a comment to mention this change tho.

pyaptamer/aptanet/tests/test_aptanet.py
```diff
@@ -81,3 +83,11 @@ def test_sklearn_compatible_estimator(estimator, check):
     Run scikit-learn's compatibility checks on the AptaNetClassifier.
     """
     check(estimator)
+
+
+def test_aptanet_layer_uses_standard_dropout():
```
satvshr: No need to test it if it is something hardcoded

satvshr:
Let's go ahead with (2) then

pyaptamer/aptanet/_aptanet_nn.py
```diff
@@ -18,7 +18,7 @@ def aptanet_layer(input_dim, output_dim, dropout, lazy=False):
         Size of each output sample (i.e., number of neurons in the layer).
 
     dropout : float
-        Dropout probability for AlphaDropout. Must be between 0 and 1.
+        Dropout probability for Dropout. Must be between 0 and 1.
```
NennoMP: Minor. Mentioning dropout two times makes the comment redundant. I would simply mention something like  `Dropout rate. Must be between 0 and 1`. I changed "probability" to "rate" to make it consistent with the description of dropout in the `AptaNetMLP` class below.

NennoMP (changes requested):

pyaptamer/aptanet/_aptanet_nn.py
NennoMP, line 47: Despite changing `AlphaDropout` to `Dropout` for good reasons... since is not in line with the original implementation I would add a comment mentioning why we are doing this. It can be a concise statement and be put here in the model description before the parameters.

kunal14901: @NennoMP I have the changes. Please have a look

siddharth7113:
@kunal14901  Your tests are still failing, please only request re-review after you have ensured tests are passing, also instead of tagging maintainer, simply re-request review here:
image

pyaptamer/aptanet/tests/test_aptanet.py
NennoMP: @kunal14901 Stavik comment is still unaddressed.

NennoMP (changes requested):

NennoMP (approved):

Issue #365, opened 2026-04-11
[BUG] pyaptamer.aptanet import fails with scikit-learn 1.4 due to validate_data


ashum9:
## **Describe the bug**

Importing `pyaptamer.aptanet` fails with certain versions of scikit-learn (e.g. 1.4+) due to usage of `validate_data`, which is not available or incorrectly imported.

This breaks the AptaNet module even though the package installs successfully.

---

## **To Reproduce**

```bash
python -m venv /tmp/test_env
source /tmp/test_env/bin/activate

pip install scikit-learn==1.4.0
pip install -e .

python - << 'EOF'
import pyaptamer.aptanet
print("Imported successfully")
EOF
```

---

## **Expected behavior**

`pyaptamer.aptanet` should import successfully across supported scikit-learn versions.

---

## **Additional context**

Current behavior results in:

```text
ImportError: cannot import name 'validate_data'
```

This suggests a mismatch between the code and the declared dependency range.

Possible fix:

* Replace usage with `self._validate_data(...)`
* OR update version constraints accordingly

---

## **Versions**

<img width="745" height="81" alt="Image" src="https://github.com/user-attachments/assets/8778cf85-d18f-415d-b4e1-b27cc4f28682" />

PR #371, opened 2026-04-11
fix: sklearn compat for validate_data import (issue #365)

mani-baktha:
`validate_data` was introduced as a module-level function in scikit-learn 1.6,
but `pyproject.toml` declares support for `scikit-learn>=1.3.0,<1.8.0`, meaning
users on 1.3–1.5 cannot use the library at all.

## Fix

Added a `try/except` fallback in `_feature_classifier.py` that imports
`validate_data` from `sklearn_compat` when the sklearn version does not expose
it as a standalone function. `sklearn_compat` is already an indirect dependency
of the project (via `imbalanced-learn`).

## Changes

- `pyaptamer/aptanet/_feature_classifier.py` — replaced bare import with
  version-safe try/except shim
- `pyaptamer/aptanet/tests/test_aptanet.py` — added regression test
  `test_validate_data_importable` to guard against future regressions

## Testing

All 106 tests pass.

### DIFF
--- a/pyaptamer/aptanet/_feature_classifier.py
+++ b/pyaptamer/aptanet/_feature_classifier.py
@@ -9,7 +9,11 @@ from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
 from sklearn.feature_selection import SelectFromModel
 from sklearn.pipeline import Pipeline
 from sklearn.utils.multiclass import type_of_target
-from sklearn.utils.validation import check_is_fitted, validate_data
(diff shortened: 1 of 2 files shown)


siddharth7113:
The [pyproject.toml](https://github.com/gc-os-ai/pyaptamer/blob/c01f3fb3ddff5810163c8fc6a7473dfe4c2bae0a/pyproject.toml#L23) has been updated.

PR #373, opened 2026-04-11
fix(rna2vec): always append padded sequence when all triplets are unknown

### LINKED ISSUE #372: `rna2vec` silently drops sequences when all triplets map to unknown tokens, breaking input/output length parity
## Problem

`rna2vec` in `pyaptamer/utils/_rna.py` drops sequences when all triplets map to `0` due to this condition:

```python
if any(converted):
```

Since `any([0, 0, ...])` evaluates to `False`, such sequences are skipped. This causes the output array to have fewer rows than the input list.

## Reproducible example

```python

out = rna2vec(["NNN", "AAAC"], sequence_type="rna", max_sequence_length=4)
print(out.shape)  # (1, 4)
# expected: (2, 4)
```

## Impact

This breaks input/output alignment. Any downstream use (e.g., pairing with labels for training) will silently misalign data, leading to incorrect results.

## Proposed fix

Remove the `if any(converted):` check and always append the padded sequence.

All-unknown sequences should produce an all-zero row, which is consistent with how unknown triplets are already mapped to `0` elsewhere in the function.

maniktyagi04:
Fixes #372 
## Summary

Remove the if any(converted): guard in rna2vec which caused sequences with all-zero values to be dropped.

This ensures the output always preserves the same number of rows as the input.

**Before:**
```python
rna2vec(["NNN", "AAAC"], sequence_type="rna", max_sequence_length=4).shape
# (1, 4)  <- "NNN" was silently dropped
```

**After:**
```python
rna2vec(["NNN", "AAAC"], sequence_type="rna", max_sequence_length=4).shape
# (2, 4)  <- all-zero row for "NNN", correct row for "AAAC"
```

### DIFF
--- a/pyaptamer/utils/_rna.py
+++ b/pyaptamer/utils/_rna.py
@@ -151,24 +151,22 @@ def rna2vec(
             triplets.get(sequence[i : i + 3], 0) for i in range(len(sequence) - 2)
         ]
 
-        # skip sequences that convert to an empty list
-        if any(converted):
-            # truncate if too long
-            if max_sequence_length is not None and len(converted) > max_sequence_length:
-                converted = converted[:max_sequence_length]
+        # truncate if too long
+        if max_sequence_length is not None and len(converted) > max_sequence_length:
+            converted = converted[:max_sequence_length]
 
-            # pad if too short
-            if max_sequence_length is not None:
-                pad_length = max_sequence_length - len(converted)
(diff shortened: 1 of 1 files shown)


siddharth7113:
See comment [here](https://github.com/gc-os-ai/pyaptamer/issues/372#issuecomment-4342411188)

PR #417, opened 2026-04-14
[BUG]: Fix `rna2vec` silent data loss for short sequences

### LINKED ISSUE #415: [BUG](violation): `rna2vec` silently drops short sequences
**Describe the bug**
`rna2vec()` in `pyaptamer/utils/_rna.py` silently skips any sequence shorter than 3 characters instead of including it as a zero-padded row. This violates the documented return shape contract of `(len(sequence_list), max_sequence_length)`

**To Reproduce**
```python

  # Single short sequence — caller expects shape (1, 275), gets (0, 275)
  result = rna2vec(["AC"], sequence_type="rna", max_sequence_length=275)
  print(result.shape)   # (0,) — empty array, NOT (1, 275)

  # Mixed list — caller expects shape (3, 275), gets (2, 275)
  result = rna2vec(["ACGU", "AC", "ACGU"], sequence_type="rna",
  max_sequence_length=275)
  print(result.shape)   # (2, 275) — one row silently missing

  The crash this causes downstream in AptamerEvalAptaTrans.evaluate():
  # rna2vec returns empty array → torch.tensor([]) → shape mismatch in the model      
  score = experiment.evaluate("AC")
  # → RuntimeError: expected input of shape [...] but got []
```

**Expected behavior**
`rna2vec` should always return an array of shape `(len(sequence_list),max_sequence_length)` as documented. Sequences too short to form a triplet should be represented as an all-zero row (i.e., treated as fully unknown/padding), matching the same zero-padding already applied to known-but-short encodings.

**Versions**
<details>
(venv) direk@dkakkar:/mnt/f/pyaptamer$ python -c "import pyaptamer; print(pyaptamer.__version__)"
0.1.0a1
(venv) direk@dkakkar:/mnt/f/pyaptamer$ 
</details>

direkkakkar319-ops:
#### Reference Issues/PRs
Fixes #415 

`rna2vec()` in `pyaptamer/utils/_rna.py` used if any(converted): to guard the append, which silently dropped any sequence shorter than 3 characters. For sequences of length `0`, `1`, or `2`, `range(len(sequence) - 2)` produces an empty list, `any([])` is False, and the sequence is never added to the result. 

Therfor , this ***violates*** the documented return shape contract of `(len(sequence_list), max_sequence_length`) — callers receive fewer rows than inputs with no error or warning, causing cryptc shape mismatches downstream in `AptamerEvalAptaTrans.`

***The fix removes the conditional guard entirely***. Every input sequence now unconditionally appends one padded row. Sequences too short to form any triplet produce `converted = []`, which zero-pads to `max_sequence_length`, preserving the shape contract.

#### PR checklist
- [x] Added/modified tests

**pytest pass**
* individuals
Image

* all
image

**pre-commits pass**
Image

### DIFF
--- a/pyaptamer/utils/_rna.py
+++ b/pyaptamer/utils/_rna.py
@@ -151,24 +151,20 @@ def rna2vec(
             triplets.get(sequence[i : i + 3], 0) for i in range(len(sequence) - 2)
         ]
 
-        # skip sequences that convert to an empty list
-        if any(converted):
-            # truncate if too long
-            if max_sequence_length is not None and len(converted) > max_sequence_length:
-                converted = converted[:max_sequence_length]
+        # truncate if too long
+        if len(converted) > max_sequence_length:
+            converted = converted[:max_sequence_length]
 
-            # pad if too short
-            if max_sequence_length is not None:
-                pad_length = max_sequence_length - len(converted)
-                padded_sequence = np.pad(
-                    array=converted,
-                    pad_width=(0, pad_length),
-                    constant_values=0,
-                )
-            else:
-                padded_sequence = np.array(converted)
+        # pad to max_sequence_length (sequences too short to form any triplet
+        # produce an empty converted list and become an all-zero row)
+        pad_length = max_sequence_length - len(converted)
+        padded_sequence = np.pad(
+            array=converted,
+            pad_width=(0, pad_length),
+            constant_values=0,
+        )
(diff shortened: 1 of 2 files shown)


siddharth7113:
See [here](https://github.com/gc-os-ai/pyaptamer/issues/284#issuecomment-4281195043)

PR #435, opened 2026-04-15
FIX: FIXED THE INFINITE CANDIDATE GENERATION

### LINKED ISSUE #434: [BUG]: Potential infinite loop in candidate generation
BUG:

AptaTransPipeline.recommend() could fail to terminate when the MCTS-based candidate generation repeatedly returned duplicate sequences. Since the loop only exited after collecting n_candidates unique items, duplicate-heavy runs could lead to extremely long runtimes or effectively infinite loops.

To Reproduce

from pyaptamer import AptaTransPipeline
pipeline = AptaTransPipeline(...)
# Configure/seed such that MCTS repeatedly proposes identical or highly similar sequences and request more unique candidates than can realistically be generated

pipeline.recommend(n_candidates=20)
# Before fix: call may run indefinitely or take excessively long to terminate

Expected behavior

recommend() should always terminate within a bounded number of iterations. If the requested number of unique candidates cannot be achieved, it should:
-Return the set of unique candidates collected so far
-Clearly notify the user (e.g., via a warning) that the uniqueness target was not met

Additional context

This issue has been resolved by introducing bounded stopping conditions:

-A maximum limit on total attempts
-A maximum limit on consecutive duplicate (stalled) attempts

If the uniqueness target is not reached within these bounds, the method now:

-Exits early
-Returns partial unique results
-Emits a RuntimeWarning to inform the user

This fix is critical for production reliability, as it prevents non-terminating calls that could block pipelines and waste computational resources.

Versions

Affected versions: All versions prior to the fix for AptaTransPipeline.recommend() termination logic (duplicate-stall handling).

Chetansahney:
Reference Issues/PRs
Fixes #434 

#What does this implement/fix?
-Fixes a bug where recommend() could hang indefinitely if MCTS kept returning duplicate candidates.
-Added bounded stopping logic (max attempts + max duplicate stalls). If uniqueness isn’t met, the method now exits early,  
 returns partial results, and emits a RuntimeWarning.

#What should reviewers focus on?
-Stopping conditions and edge cases
-Early return behavior and output correctness
-Warning emission

#Did you add tests?
Yes - covered duplicate stalls, early termination, warnings, and partial results.

#Any other comments?
Prevents non-terminating calls and improves production reliability.

#PR checklist:
-[BUG] prefix used
-Tests added/updated
-Pre-commit checks passed

### DIFF
--- a/pyaptamer/aptatrans/_pipeline.py
+++ b/pyaptamer/aptatrans/_pipeline.py
@@ -6,6 +6,7 @@ candidate aptamers recommendation.
 __author__ = ["nennomp"]
 __all__ = ["AptaTransPipeline"]
 
+import warnings
 import torch
 from torch import Tensor
 
@@ -86,6 +87,10 @@ class AptaTransPipeline:
     >>> candidates = pipeline.recommend(target, n_candidates=1, verbose=False)
     """
 
+    _STALLED_ATTEMPTS_PER_CANDIDATE = 3
+    _MIN_STALLED_ATTEMPTS_THRESHOLD = 10
+    _MAX_ATTEMPTS_PER_CANDIDATE = 10
+    
     def __init__(
         self,
         device: torch.device,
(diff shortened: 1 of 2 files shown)


direkkakkar319-ops (commented):
hi i think your pr is fixing similar issue like of #385

PR #441, opened 2026-04-17
[ENH] Refactor APIDataset to inherit BaseObject ; unify AptaTrans and AptaNet input handling

### LINKED ISSUE #430: [ENH] Moving `APIDataset` to inherit from a base dataset object
Currently the [dataclass](https://github.com/gc-os-ai/pyaptamer/tree/main/pyaptamer/datasets/dataclasses) is only defined to inherit from torch, and performs multiple preprocessing step, it would be nice to decouple them and generalise to inherit from a base class.

More details: TBD

**Additional context**

siddharth7113:
#### Reference Issues/PRs

Fixes #430 . Coupled with PR #231

> AI disclousre: Parts of code and description has been generated with Opus 4.7 but verified manually.

Refactors `APIDataset` from an encoding-baked-in torch `Dataset` subclass into a pure in-memory paired-sequence container, so that both `AptaTransPipeline` and `AptaNetPipeline` can consume the same data container in any of several supported input shapes.

**The problem:** Before this PR, `APIDataset` mixed three concerns (data containment, encoding, torch integration) and was usable only by AptaTrans. AptaNet had its own input convention (`list[tuple]`). Adding a third algorithm would require a third convention.

**The fix:** Three-layer split.

```
Layer 1: MoleculeLoader              — file I/O (unchanged)
Layer 2: BaseAptamerDataset          — NEW abstract root
         ├── APIDataset              — pure container (raw strings)
         └── MaskedDataset           — migrated, real subclass
Layer 3: Pipelines + encoding        — AptaTrans/AptaNet pipelines call
                                       existing utility functions at their own seam
```

**New class hierarchy:**

```python
class BaseAptamerDataset(BaseObject):
    """Abstract root — tag plumbing + load() contract."""
    _tags = {"object_type": "dataset", "scitype": None, "X_inner_mtype": None, "has_y": True}
    def load(self): raise NotImplementedError

class APIDataset(BaseAptamerDataset):
    _tags = {"scitype": "APIPairs",
             "X_inner_mtype": ["pd.DataFrame", "list_tuples", "numpy_arrays"]}

class MaskedDataset(BaseAptamerDataset, torch.utils.data.Dataset):
    _tags = {"scitype": "MaskedSequences", "X_inner_mtype": ["numpy_arrays_pair"]}
```

**`APIDataset.from_any(X, y)` accepts five input shapes:**

```python
# any of these work for both pipelines:
pipe.fit(list_of_tuples, y)
pipe.fit(dataframe, y)                     # columns: aptamer, protein
pipe.fit((numpy_apta, numpy_prot), y)
pipe.fit((molecule_loader_a, molecule_loader_b), y)
pipe.fit(api_dataset, api_dataset.y)       # pass-through
```

**AptaTrans pipeline gets a new internal helper:**

```python
def _prepare_dataloader(self, X, y=None, train=False, batch_size=32):
    ds = APIDataset.from_any(X, y)
    df = ds.load()
    x_apta_enc = rna2vec(df["aptamer"].tolist(), max_sequence_length=self.model.apta_embedding.max_len, sequence_type="rna")
    x_prot_enc = encode_rna(df["protein"].tolist(), words=self.prot_words, max_len=self.model.prot_embedding.max_len)
    torch_ds = _AptaTransTorchDataset(x_apta_enc, x_prot_enc, ds.y)
    return DataLoader(torch_ds, batch_size=batch_size, shuffle=train)
```

This is the seam for the stacked follow-up PR that adds sklearn-style `fit(X, y)` / `predict(X)`. In this PR, `AptaTrans`'s public interface (`predict(candidate, target)`, `recommend(...)`) is **unchanged**.

**Breaking changes:**

- `APIDataset` constructor: removed `apta_max_len`, `prot_max_len`, `prot_words`, `split` kwargs. These were encoding params that don't belong on a data container.
  - `apta_max_len` / `prot_max_len` now read from `self.model.apta_embedding.max_len` / `self.model.prot_embedding.max_len`
  - `prot_words` is already on `AptaTransPipeline.prot_words`
  - `split="train"` → `train=True` on `_prepare_dataloader` (just controls shuffle; augmentation is now the user's job)

- `APIDataset` no longer inherits `torch.utils.data.Dataset`. Code that did `DataLoader(api_dataset, ...)` directly breaks. Use `pipeline._prepare_dataloader(api_dataset, ...)` instead.

- `APIDataset` no longer does label encoding internally. Users with string labels (`"positive"`/`"negative"`) must encode to `0`/`1` before passing to the pipeline.

- `APIDataset` no longer does data augmentation internally. Users must call `augment_reverse` explicitly before constructing the dataset.

**Tutorial updates:** `examples/aptatrans_tutorial.ipynb` cell 10 now does explicit label encoding + augmentation, and a new cell after model initialization creates DataLoaders via `pipeline._prepare_dataloader(...)`. `examples/aptanet_tutorial.ipynb` needs no changes (list-of-tuples path still works).

- [ ] **Design:** is `BaseAptamerDataset` as an abstract root (only tags + `load()`) the right call, vs. something richer?
- [ ] **Breaking changes:** is a clean break on `APIDataset`'s old kwargs acceptable for early alpha, or do we want a deprecation cycle?
- [ ] **Mtype registry scope:** `_mtype.py` is APIPairs-only with a documented path to add sibling modules. Acceptable for now?
- [ ] **Seam design:** `_prepare_dataloader` as the hook point for the stacked `fit`/`predict` PR — is the seam in the right place?

 **248 new tests added** .

- `tests/test_mtype.py` (13 tests): mtype dispatcher round-trips, coercion, unsupported input errors
- `tests/test_base.py` (4 tests): abstract root tag contract, `load()` NotImplementedError
- `tests/test_api.py` (19 tests): construction with all input shapes, `from_any` pass-through, y coercion (DataFrame/2D → 1D), breaking-change validation
- `tests/test_masked.py` (5 tests): inheritance, new `load()` returning `(x, y)`, existing behavior preserved
- `tests/test_torch_dataset.py` (6 tests): per-sample tensor cast, DataLoader compat, no augment param
- `tests/test_aptatrans.py`: 2 new tests for `_prepare_dataloader` across input shapes
- `tests/test_aptanet.py`: 4 new parametrized tests covering DataFrame, numpy pair, APIDataset, and DataFrame + string-label inputs (the real `load_li2014` flow)

#### PR checklist

- [x] The PR title starts with `[ENH]`
- [x] Added/modified tests
- [x] Used pre-commit hooks — all files pass ruff-format, ruff, mixed-line-ending


pyaptamer/datasets/dataclasses/_base.py
```diff
@@ -0,0 +1,49 @@
...
+          (e.g. ``"APIPairs"``, ``"MaskedSequences"``).
+        - Override `_tags["X_inner_mtype"]` with the canonical inner mtype(s)
+          their `load()` returns.
+        - Implement `load()` to return data in their canonical inner mtype.
+
+    The base provides no input-coercion logic — what counts as a valid input
+    depends entirely on the scitype, so coercion lives in the concrete subclass.
+    """
+
+    _tags = {
```
deveshsy: @siddharth7113 

Reviewing the new BaseAptamerDataset abstraction moving to skbase.base.BaseObject and using a tag system is a massive upgrade for interoperability. It makes the whole pipeline much more agnostic, which is great for swapping between deep learning models and classical ML like Random Forest without a total rewrite.

I was digging through the base tags in base.py and had a quick thought regarding "has_y": True being hardcoded at the abstract root. This works perfectly for the training phase, but I’m curious how it plays with inference/scoring workflows where y is naturally None.

If a user passes raw sequences into predict(X) using this new container, will the dataset get grumpy because it's looking for a target that isn't there?

Maybe it’s worth making has_y dynamic or setting it to False by default in the base, then flipping it to True only when y is actually passed during from_any() initialization? Curious to hear your take on whether you'd rather handle that check here or further down in the pipeline logic.

DashratRajpurohit:
Hi @siddharth7113, I’ve been reviewing the new `APIDataset` refactor and found a runtime blocker in the torch dataset implementation.

Specifically, in` _AptaTransTorchDatase`t, returning None for labels during inference crashes the` DataLoader`'s default collator. Here is a minimal script to reproduce the crash:

```
import torch
from torch.utils.data import DataLoader, Dataset
class MockDataset(Dataset):
    def __init__(self):
        self.x = [torch.randn(10) for _ in range(5)]
        self.y = None # Unlabeled inference
    def __len__(self): return len(self.x)
    def __getitem__(self, i):
        # Current logic in PR #441
        if self.y is None: return self.x[i], None
        return self.x[i], self.y[i]
loader = DataLoader(MockDataset(), batch_size=2)
for batch in loader: # Throws TypeError
    print(batch)
```
This throws: TypeError: batch must contain tensors, numpy arrays, numbers, dicts or lists; found <class 'NoneType'>.

**I also noticed that the `ignore_duplicates `flag from my refactor in #443 isn't yet being passed through in from_any. I'm happy to provide a patch for both of these if you'd like!**

pyaptamer/datasets/dataclasses/_api.py
```diff
@@ -1,110 +1,121 @@
...
-        x_prot = encode_rna(
-            sequences=x_prot,
-            words=self.prot_words,
-            max_len=self.prot_max_len,
-        )
-        y = torch.tensor((y == "positive").astype(int))
+        if isinstance(X, APIDataset):
+            return X
+        df = coerce_input(X)
+        return cls(x_apta=df["aptamer"], x_prot=df["protein"], y=y)
```
aditi-dsi, line 109: @siddharth7113, these `"aptamer"` and `"protein"` column names are hardcoded in the `_mtype` and `_api` functions. Since the goal of the PR is to improve dataset generalization, should we make these column names dynamic? 

For example, what if we just add `apta_col` and `prot_col` kwargs to `APIDataset.from_any()`? That way, we can dynamically rename the user's columns right at the entry point, keeping the internal code as is.

examples/aptatrans_tutorial.ipynb
```diff
@@ -255,50 +238,55 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
satvshr, line 241: Something to look into?

pyaptamer/aptanet/tests/test_aptanet.py
```diff
@@ -70,6 +70,66 @@ def test_pipeline_fit_and_predict_regression(aptamer_seq, protein_seq):
     assert np.issubdtype(preds.dtype, np.floating)
 
 
+@pytest.mark.parametrize("aptamer_seq, protein_seq", params)
```
satvshr: I would recommend not adding additional tests and increasing the scope of this PR (covers 19 files now), I'd only keep the APIDataset test as it is relevant to this PR, I'd move the other tests to a newer issue/PR

pyaptamer/aptanet/_pipeline.py
```diff
@@ -13,8 +13,7 @@
 
 
 class AptaNetPipeline(BaseObject, BaseEstimator):
-    """
-    AptaNet algorithm for aptamer–protein interaction prediction [1]_
+    """AptaNet algorithm for aptamer-protein interaction prediction [1]_.
```
satvshr, line 16: Is this ruff making these changes? I am confused as to what is going on here

pyaptamer/aptanet/_pipeline.py
```diff
@@ -76,14 +83,28 @@ def _build_pipeline(self):
...
         check_is_fitted(self)
-        return self.pipeline_.predict_proba(X)
+        pairs, _ = self._to_pairs(X)
+        return self.pipeline_.predict_proba(pairs)
 
     def predict(self, X):
         check_is_fitted(self)
-        return self.pipeline_.predict(X)
+        pairs, _ = self._to_pairs(X)
+        return self.pipeline_.predict(pairs)
```
satvshr, line 110: It seems like we are forcing aptanet to use APIDataset forcibly internally? I am confused as to why we are doing that

pyaptamer/aptatrans/_torch_dataset.py
```diff
@@ -0,0 +1,54 @@
...
+
+__author__ = ["siddharth7113"]
+__all__ = ["_AptaTransTorchDataset"]
+
+import numpy as np
+import torch
+from torch.utils.data import Dataset
+
+
+class _AptaTransTorchDataset(Dataset):
```
satvshr: Is this the original `APIDataset` class?

pyaptamer/datasets/dataclasses/tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+"""Shared test data for dataclasses tests.
```
satvshr, line 1: Is this really required for 2 files? I mean, it is just 4 lines of variables

pyaptamer/datasets/dataclasses/_base.py
```diff
@@ -0,0 +1,49 @@
...
+
+class BaseAptamerDataset(BaseObject):
+    """Abstract root for pyaptamer in-memory data containers.
+
+    Subclasses MUST:
+        - Override `_tags["scitype"]` with their scientific-type identifier
+          (e.g. ``"APIPairs"``, ``"MaskedSequences"``).
+        - Override `_tags["X_inner_mtype"]` with the canonical inner mtype(s)
+          their `load()` returns.
+        - Implement `load()` to return data in their canonical inner mtype.
```
satvshr: Can't we have a small test to ensure the output from `load()` and `_tags["X_inner_mtype"]` matches?

pyaptamer/datasets/dataclasses/_masked.py
```diff
@@ -59,6 +61,14 @@ class MaskedDataset(Dataset):
     2
     """
 
+    _tags = {
+        "scitype": "MaskedSequences",
+        # Descriptive only — no mtype dispatcher registered for this scitype yet.
+        # Add _mtype_masked.py sibling when conversion is needed.
+        "X_inner_mtype": ["numpy_arrays_pair"],
```
satvshr: Base class (`BaseAptamerDataset`) should also check if these types are valid no?

pyaptamer/datasets/dataclasses/_mtype.py
```diff
@@ -0,0 +1,157 @@
+"""Mtype dispatcher for the ``APIPairs`` scitype.
+
+Internal module. Defines the supported mtypes for paired aptamer-protein
+data and a ``convert_to`` function that converts between them. Used by
+``APIDataset`` for input coercion; not part of the public API.
+
+**Scitype scope:** this module handles the ``APIPairs`` scitype only.
+Other scitypes (``MaskedSequences``, future ones) maintain their own mtype
```
satvshr: If it is based on the scitype then why is the file name `pyaptamer/datasets/dataclasses/_mtype.py` and not `pyaptamer/datasets/dataclasses/api_pairs.py`, also, where is the mtype conversion file for `MaskedSequences`? Shouldn't it have a `convert_to` function too? I am a bit confused

satvshr (commented):
Nice! I'm working to understand the design behind these changes. I have dropped a few comments based on what I see so far; once I review the design, I will add more comments and questions.

deveshsy (commented):
@siddharth7113 I’ve been tracing the APIDataset initialization logic here. I noticed that while the init correctly handles y=None for the data attribute, it doesn't override the inherited has_y: True tag from the base class.

Since _tags is a static class attribute in BaseAptamerDataset, any instance created for inference still reports has_y: True. This 'metadata lie' is what triggers the TypeError Dashrat found—the DataLoader sees the tag, expects labels, and attempts to batch None objects.

A clean fix would be to add self.set_tags(has_y=(y is not None)) in the constructor. This keeps the metadata in sync with the instance data without needing to modify the base class.

DashratRajpurohit:
Looked into this a bit more after` deveshsy’s` note about the metadata tags. For now, until we fix the inheritance properly, we can just avoid returning None in `_torch_dataset.py` to stop the DataLoader from crashing.

If y isn’t there, we can return a dummy tensor like` torch.tensor(0.0)` so batching doesn’t break.

```
y = self.labels[idx] if self.labels is not None else 
torch.tensor(0.0, dtype=torch.float32)
return x, y
```

Not a proper fix, but it should keep things from crashing fro now. We can clean it up once the` has_y` logic is sorted.

DashratRajpurohit:
I've looked into this further with deveshsy's findings about the metadata tags. While we fix the tag inheritance, a safe 'hotfix' for the DataLoader crash in _torch_dataset.py is to ensure we never return None. 

By returning a dummy tensor (e.g., torch.tensor(0.0)) when y is missing, we satisfy the PyTorch batching requirement without crashing the pipeline. 

```python
# Proposed fix in _AptaTransTorchDataset.__getitem__
y = self.labels[idx] if self.labels is not None else torch.tensor(0.0, dtype=torch.float32)
return x, y
```

This keeps the API implementation clean until the higher-level has_y tag logic is fully refactored. Happy to open a PR against your branch with this if you'd like!

pyaptamer/aptanet/_pipeline.py
siddharth7113: I am not sure, something weird is happening after I run precommit it shows no changes but some docstring changes like this show up, it could be artefact from using AI too, but I am not sure 100%

pyaptamer/aptanet/_pipeline.py
siddharth7113: I hope now this makes more sense. 😅

pyaptamer/aptatrans/_torch_dataset.py
siddharth7113: Yes and No, this was written for converting the `APIDataset` to Tensor format for loading, but I don't think we need it there is something called `TensorDataset` in pytorch that does exactly did and this can be part of `_prepare_dataloader` in AptaTransPipeline.

pyaptamer/aptanet/tests/test_aptanet.py
siddharth7113: These tests check whether the APIDataset is accepted by `AptaNet` in different formats.

pyaptamer/datasets/dataclasses/tests/conftest.py
siddharth7113: I would say let's leave it here, since this is common enough to be used by other tests as well in future instead of each file writing strings.

pyaptamer/datasets/dataclasses/_api.py
siddharth7113: Good catch! will update it to have args to be passed for column names

pyaptamer/aptanet/tests/test_aptanet.py
siddharth7113: but I think 1 test should do the job too, this is just too broad test coverage.

PR #450, opened 2026-04-19
[ENH] Clarify MCTS simulation length formula and add invariant test

### LINKED ISSUE #449: [BUG] MCTS simulation length formula obscures invariant, missing length-invariant test
**Describe the bug**

_simulation in pyaptamer/mcts/_algorithm.py contains:                                                     
                                                                                                                
  ```python
remaining_length = self.depth - (len(sequence) // 2) 
```                                                       
                                                                                                                
The formula is mathematically correct — each encoded state is a 2-char pair, len(sequence) // 2 is the current nucleotide count, and the loop correctly fills to self.depth total nucleotides. However:                     
                                                                                                              
  1. The formula's intent is non-obvious: it isn't clear that len(sequence) // 2 already accounts for both      
  self.base and the tree path, leading to the misread that self.base is "double-counted."
  2. The existing test_simulation only checks that the method does not raise — it never asserts that the        
  evaluated sequence has the right length. 

**To Reproduce**
 No runtime failure; the risk is a silent regression if the formula is ever edited under the wrong mental  model. 

**Expected behavior**
  _simulation should always pass a sequence of exactly depth nucleotides to experiment.evaluate, regardless of
  the node's depth in the tree and regardless of how many nucleotides self.base already contains.               

**Versions**
0.1.0a1

### PR DESCRIPTION
**Pull request by ritankarsaha**
#### Reference Issues/PRs

Fixes #449 

- Renames remaining_length's computation to use nucleotides_so_far as an intermediate variable 
  - Adds test_simulation_sequence_length covering four cases: empty base at shallow depth, empty base at
  near-terminal depth, non-empty base (simulating a later round), and calling simulation directly on the tree root.

  The existing test_simulation only checked that the method ran without crashing. There was no test asserting
  the evaluated sequence always has exactly depth nucleotides, leaving the formula unguarded against future
  misedits.

  - pytest pyaptamer/mcts/tests/test_mcts.py — all 37 tests pass
  - pre-commit run --all-files — clean

- [x] Added/modified tests

### DIFF
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -223,8 +223,10 @@ class MCTS(BaseObject):
         # prepend the `self.base` sequence
         sequence = self.base + sequence
 
-        # fill the rest of the sequence with random possible values
-        remaining_length = self.depth - (len(sequence) // 2)
+        # Each encoded state is a 2-char pair (e.g. "A_", "_C") representing one
+        # nucleotide
+        nucleotides_so_far = len(sequence) // 2
+        remaining_length = self.depth - nucleotides_so_far
         for _ in range(remaining_length):
             sequence += random.choice(self.states)
 
@@ -245,7 +247,8 @@ class MCTS(BaseObject):
         curr = self.root
         subsequence = self.base
 
-        # traverse the tree
+        # Number of tree levels still needed to reach `self.depth` nucleotides total.
+        # len(self.base) // 2 is the nucleotide count already committed to self.base.
         max_steps = self.depth - (len(self.base) // 2)
         for _ in range(max_steps):
             if not curr.children:
--- a/pyaptamer/mcts/tests/test_mcts.py
+++ b/pyaptamer/mcts/tests/test_mcts.py
@@ -309,6 +309,40 @@ class TestMCTS:
         _ = mcts._simulation(node=node)
         assert True
 
+    def test_simulation_sequence_length(self, mcts):
+        """Simulation must always evaluate a sequence of exactly `depth` nucleotides."""
+        evaluated = []
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by ritankarsaha** [comment]
@siddharth7113 @NennoMP can you please take a look into this PR ?

File: pyaptamer/mcts/tests/test_mcts.py [hunk 1]
```diff
@@ -309,6 +309,40 @@ def test_simulation(self, mcts):
...
+        node_r2 = mcts.root.create_child(val="_U")
+        mcts._simulation(node_r2)
+
+        mcts._simulation(mcts.root)
+
+        assert all(length == mcts.depth for length in evaluated), (
+            f"Simulation evaluated sequences of wrong length: {evaluated} "
+            f"(expected all {mcts.depth})"
+        )
+
```
NennoMP, line 345: Whether the length of generated candidates matches depth is implicitly checked in `test_find_best_subsequence`, admittely only for the case of `depth=5`.

pyaptamer/mcts/_algorithm.py
```diff
@@ -223,8 +223,10 @@ def _simulation(self, node: "TreeNode") -> float:
         # prepend the `self.base` sequence
         sequence = self.base + sequence
 
-        # fill the rest of the sequence with random possible values
-        remaining_length = self.depth - (len(sequence) // 2)
+        # Each encoded state is a 2-char pair (e.g. "A_", "_C") representing one
+        # nucleotide
+        nucleotides_so_far = len(sequence) // 2
+        remaining_length = self.depth - nucleotides_so_far
```
NennoMP, line 229: I really don't get this change. If the purpose is to make cleared this step the most reasonable thing to do would be to enhance the docstrings, not adding a new variable.

PR #459, opened 2026-04-19
[BUG] Clarify MCTS simulation length invariant

### LINKED ISSUE #449: [BUG] MCTS simulation length formula obscures invariant, missing length-invariant test

NandiniDhanrale:
#### Reference Issues/PRs
  Fixes #449

  This PR clarifies the sequence-length invariant in `MCTS._simulation()` without changing behavior.

  The calculation that determines how many random states to append is now split into an explicit `current_length`
  intermediate value, making it clear that `len(sequence) // 2` counts encoded nucleotide pairs.

  - Whether the renamed intermediate variable makes the invariant easier to understand.
  - Whether the regression test correctly protects the expected simulation length behavior.

  Yes.
  - Added a regression test that captures the sequence passed to `experiment.evaluate()`.
  - The test asserts that the evaluated encoded sequence has the expected total length and that the decoded nucleotide
  count matches `self.depth`.

  #### Any other comments?
  This is a behavior-preserving refactor with a focused regression test. It is intended to make the invariant harder to
  misread in future edits.

  #### PR checklist
  - [x] Added/modified tests

### DIFF
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -223,8 +223,12 @@ class MCTS(BaseObject):
         # prepend the `self.base` sequence
         sequence = self.base + sequence
 
+        # `sequence` stores encoded nucleotide pairs, so two characters represent
+        # one nucleotide.
+        current_length = len(sequence) // 2
+
         # fill the rest of the sequence with random possible values
-        remaining_length = self.depth - (len(sequence) // 2)
+        remaining_length = self.depth - current_length
         for _ in range(remaining_length):
             sequence += random.choice(self.states)
 
--- a/pyaptamer/mcts/tests/test_mcts.py
+++ b/pyaptamer/mcts/tests/test_mcts.py
@@ -309,6 +309,29 @@ class TestMCTS:
         _ = mcts._simulation(node=node)
         assert True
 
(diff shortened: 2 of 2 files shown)


ritankarsaha:
Hey there is already a PR for this which is already partially approved ! 😅 @NandiniDhanrale

PR #468, opened 2026-04-20
[BUG] fix AptaNetPipeline.fit to return self, fixing sklearn method chaining

### LINKED ISSUE #467: [BUG] AptaNetPipeline.fit() does not return self, breaking sklearn estimators contract and method chaining
In `pyaptamer/aptanet/_pipeline.py`, the `fit()` method for `AptaNetPipeline` does not return `self`.

Per the Scikit-learn estimator contract, `fit()` must return the fitted estimator. Because it implicitly returns `None`, it breaks method chaining (e.g., `pipe.fit(X, y).predict(X)` raises an `AttributeError`) and prevents the pipeline from being used properly inside standard sklearn utilities like `GridSearchCV` or `cross_validate`.

**To Reproduce**

```python
from pyaptamer.aptanet import AptaNetPipeline

pipe = AptaNetPipeline()
X, y = [("ACG", "AC")], [0]
pipe.fit(X, y).predict(X) 
# AttributeError: 'NoneType' object has no attribute 'predict'
```
The fit() method should end with return self:
```python
    def fit(self, X, y):
        self.pipeline_ = self._build_pipeline()
        self.pipeline_.fit(X, y)
        return self
```
 I ran into this while trying to wrap AptaNetPipeline in a hyperparameter tuning grid, where sklearn expects the estimator instance to be returned after fitting. A one-line addition fixes this completely.

rupeshca007:
#### Reference Issues/PRs
Fixes #467

`AptaNetPipeline` inherits from `sklearn.base.BaseEstimator`, which strictly mandates that `fit()` must return `self`. Previously, `fit()` returned `None`. 

This PR simply adds `return self` to `AptaNetPipeline.fit()`. 

Without this fix, method chaining (e.g. `pipeline.fit(X, y).predict(X)`) crashes with an `AttributeError`, and wrapping this pipeline in standard wrappers (like hyperparameter grid search or cross-validation) fails when the utility expects the fitted estimator back.

That the estimator API contract is now fully satisfied.

- [ ] No explicit tests added since it's a standard API compliance fix.

#### PR checklist
- [ ] Added/modified tests

### DIFF
--- a/pyaptamer/aptanet/_pipeline.py
+++ b/pyaptamer/aptanet/_pipeline.py
@@ -79,6 +79,7 @@ class AptaNetPipeline(BaseObject, BaseEstimator):
     def fit(self, X, y):
         self.pipeline_ = self._build_pipeline()
         self.pipeline_.fit(X, y)
+        return self
 
     def predict_proba(self, X):
         check_is_fitted(self)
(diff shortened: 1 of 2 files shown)


satvshr:
Look at tests please

satvshr:
Thanks!

PR #477, opened 2026-04-21
[ENH] feat :replacing `print()`  statements with Python logging module.

### LINKED ISSUE #416: [ENH] Replace print() statements with Python logging module

The codebase uses `print()` statements throughout for debug/progress output 
(found in `mcts/_algorithm.py`, `aptatrans/_pipeline.py`, `aptatrans/_model.py`, 
`pseaac/_pseaac_aptanet.py`, `pseaac/_pseaac_general.py`, and `utils/_rna.py`).
This makes it impossible for library consumers to control log verbosity 
programmatically and clutters stdout in production use.

1.  Add a module-level logger to each affected file:
    ```python
    import logging
    logger = logging.getLogger(__name__) 

2. Replace print(...) calls with appropriate `logger.info(...)`, `logger.debug(...)`, or `logger.warning(...)`.
3. Preserve existing verbose parameters , when verbose=False, skip the log calls (or use `logger.debug` level).
4. Do NOT add any logging configuration (no handlers/formatters), leave that to the library consumer.

- Using `tqdm` for progress bars (complementary, not a replacement for logging)
- `warnings `module (only appropriate for warnings, not general debug output)

**Additional context**

Files to modify:

1. pyaptamer/mcts/_algorithm.py
2. pyaptamer/aptatrans/_pipeline.py
3. pyaptamer/aptatrans/_model.py
4. pyaptamer/pseaac/_pseaac_aptanet.py
5. pyaptamer/pseaac/_pseaac_general.py
6. pyaptamer/utils/_rna.py

ChandanKT-git:
#### Reference Issues/PRs
Fixes #416.

This PR migrates the codebase from using `print()` statements for progress and debug reporting to the standard Python `logging` module. This allows library consumers to control log verbosity programmatically and prevents cluttering `stdout` in production environments.

**Key Changes:**
*   **Module-level Loggers:** Added `logger = logging.getLogger(__name__)` to `mcts/_algorithm.py`, `aptatrans/_pipeline.py`, and `aptatrans/_model.py`.
*   **Logging levels:** 
    *   Used `logger.debug()` for high-frequency MCTS round updates and internal search progress.
    *   Used `logger.info()` for lifecycle events like model weight loading, downloads, and final candidate recommendations.
*   **Performance Optimization:** Implemented `logger.isEnabledFor(logging.DEBUG)` guards in the MCTS hot loop (`run` method) to avoid unnecessary string formatting overhead when debug logging is disabled.
*   **Ruff Enforcement:** Updated `pyproject.toml` to include the `T20` (flake8-print) rule set in the Ruff configuration. This will cause linting to fail if `print()` or `pprint()` statements are accidentally reintroduced in the future.

*   Verify that the chosen log levels (`DEBUG` vs `INFO`) are appropriate for the respective output.
*   Check the implementation of the performance guards in `pyaptamer/mcts/_algorithm.py`.

The changes were verified by running `ruff check pyaptamer/ --select T20` to ensure that no `print()` statements remain in the library code and that the new linting rule is active. Syntax correctness was verified with a full AST parse check of the modified files.

#### PR checklist
- [ ] Added/modified tests


satvshr:
@siddharth7113 @NennoMP what do you feel about our first global variable, for logging? Doesnt make sense to intialize a variable and use it for every file

ChandanKT-git:
I followed the standard Python logging pattern, but I'm happy to refactor it based on your recommendation. 

Would you prefer a centralized logger (e.g., in `pyaptamer/__init__.py`) that we use everywhere, or shall I just call wherever it's needed eg. `logging.getLogger(__name__).info("msg")`?

ChandanKT-git:
> I followed the standard Python logging pattern, but I'm happy to refactor it based on your recommendation.
> 
> Would you prefer a centralized logger (e.g., in `pyaptamer/__init__.py`) that we use everywhere, or shall I just call wherever it's needed eg. `logging.getLogger(__name__).info("msg")`?

The downside with centralised logger is that all logs will be tagged as pyaptamer instead of the specific file name (e.g., pyaptamer.mcts).

satvshr:
Could you look into how [openml](https://github.com/openml/openml-python/blob/main/openml/_config.py) does it?

ChandanKT-git:
They define a central logger for the whole library, also they have config manager class, shall I do the same create `_config.py` , that just handles logging for now, if you feel tracing of specific file is not that important, a simple centralised logger + importing it in submodules will be sufficient.

satvshr:
yes

ChandanKT-git:
> Could you look into how [openml](https://github.com/openml/openml-python/blob/main/openml/_config.py) does it?

I have implemented the centralised logging  and any other configuration you want me to implement. ?

satvshr:
Did you ask AI to copy paste the openml setup for this repository? If yes, kindly don't, this PR only needs a logger, I don't know why there is a lot of boilerplate code, we don't need a config class, just a logger class for now

ChandanKT-git:
> Did you ask AI to copy paste the openml setup for this repository? If yes, kindly don't, this PR only needs a logger, I don't know why there is a lot of boilerplate code, we don't need a config class, just a logger class for now 

No, I asked you in the above messages they have a config manager shall I do the same for pyaptamer, you said yes. Thats why I proceeded. 

Also the need I felt of this is in order to avoid losing the context of from which func/file  logging comes as centralised logger logs under tag of pyaptamer, instead of pyaptamer.mct (the specific file)

I hope I'm making sense, if not then will just implement the logger class without worrying about traces.

satvshr:
> Also the need I felt of this is in order to avoid losing the context of from which func/file logging comes as centralised logger logs under tag of pyaptamer, instead of pyaptamer.mct (the specific file)

Could you give an example? More detailed given you are giving me options to choose from but not what each option entails, if that makes sense.

ChandanKT-git:
If we just create a single logger instance (e.g., logger = logging.getLogger("pyaptamer")) and import that exact same variable into every file, we destroy the Python logging hierarchy.

Example:

```python
# Inside mcts/_algorithm.py
logger.error("Search failed due to timeout.")
# Inside aptatrans/_pipeline.py
logger.info("Data loaded successfully.")
```
Output: 
```text
[ERROR] [pyaptamer] Search failed due to timeout.
[INFO] [pyaptamer] Data loaded successfully.
```
**The Problem**: if you see an error trace from [pyaptamer], you have absolutely no idea which of the 20+ files in the repository actually threw the error. Debugging becomes a nightmare because we lost the __name__ namespace.

OpenML uses their OpenMLConfigManager class at centrally defines the handlers, file outputs, and verbosity levels. And in children file they still have global variable. [See here](https://github.com/openml/openml-python/blob/main/openml/datasets/dataset.py#L25)

I thought you wanted that way so I implemented it.

**Another option is let's say it Option B:** 
To know which file a log came from, we have to rely on the formatter using %(module)s instead of the logger's name.

Example implementation: In pyaptamer/_logger.py:

```python
import logging
# Create the one global logger
logger = logging.getLogger("pyaptamer")
logger.setLevel(logging.INFO)
# Format it to show the calling module so we know where logs come from
formatter = logging.Formatter('[%(levelname)s] [%(module)s] %(message)s')
handler = logging.StreamHandler()
handler.setFormatter(formatter)
logger.addHandler(handler)
```

In submodules (e.g., mcts/_algorithm.py):

```python
# We just import the global variable directly. No initialization needed.
from pyaptamer._logger import logger
logger.info("Running MCTS...") 
# Output: [INFO] [_algorithm] Running MCTS..
```

I hope I'm making things clear for you now, shall I go with Option B?

satvshr:
Perfect explanation! Thanks for the clarity! Let's go with option B for now given we have nothing to add to config for now, hence we do not need a config similar to openml.

ChandanKT-git:
> Perfect explanation! Thanks for the clarity! Let's go with option B for now given we have nothing to add to config for now, hence we do not need a config similar to openml.

Glad! Its done, and ready for review and merge.

pyaptamer/mcts/_algorithm.py
```diff
@@ -296,8 +299,8 @@ def run(self, verbose: bool = True) -> dict:
         # continue until we reach the target sequence length (i.e, depth * 2)
         round_count = 0
         while len(self.base) < self.depth * 2:
-            if verbose:
-                print(f"\n ----- Round: {round_count + 1} -----")
+            if verbose and logger.isEnabledFor(logging.DEBUG):
```
satvshr, line 302: Why the `and`?

ChandanKT-git: It's a performance guard.

The key reason to use **`and`** with `isEnabledFor` explicitly rather than just calling `logger.debug(...)` directly is to avoiding expensive argument computation. 

As we know Python evaluates function arguments before the function runs.

satvshr: How expensive is it? We do not expect future contributors to use this, could you compare the performances and let me know?

satvshr:
Code quality checks are failing, kindly use `pre-commit` in your env before pushing changes.

pyaptamer/mcts/_algorithm.py
ChandanKT-git: In this file the logging arguments are cheap, so the gain is small per call, but it avoids repeated formatting and any future expensive debug expressions inside the loop. 
If you want simpler code, I could remove the guard and rely on `logger.debug(..)`
So it will be easy on future contributors

satvshr: if the improvement is a microsecond it is not important, the problem would have been adding it everywhere in the future. Anyways, keep it, code quality is still failing though, do you not have `precommit` set up?

ChandanKT-git: As of now yes the improvement is a microsecond 😅, I thought I have set up the pre-commit but checks are telling me other wise. I would have to check it once again.

ChandanKT-git:
> Code quality checks are failing, kindly use `pre-commit` in your env before pushing changes.

The checks have been passing now in my ENV.

satvshr (approved):
LGTM

PR #478, opened 2026-04-21
Update docstring for augment_reverse function

### LINKED ISSUE #335: [ENH] augment_reverse() computes reverse but docstrings mentions `complement` which should be updated.

siddharth7113:
Resolves #335

### DIFF
--- a/pyaptamer/utils/_augment.py
+++ b/pyaptamer/utils/_augment.py
@@ -5,7 +5,7 @@ import numpy as np
 
 
 def augment_reverse(*sequence_arrays: np.ndarray) -> tuple[np.ndarray, ...]:
-    """Augment arrays of sequences by adding their reverse complement.
+    """Augment arrays of sequences by adding the reversed sequences.
 
     Parameters
     ----------
@@ -15,7 +15,7 @@ def augment_reverse(*sequence_arrays: np.ndarray) -> tuple[np.ndarray, ...]:
     Returns
     -------
     tuple[np.ndarray, ...]
-        A tuple of arrays, each containing sequences with their reverse complements
+        A tuple of arrays, each containing sequences with their reversed sequences.
         added.
     """
     results = []


ChandanKT-git:
Hey @siddharth7113, we need to update in dataclasses `_api.py` as well.

siddharth7113:
Can you clarify @ChandanKT-git what does this mean?

ChandanKT-git:
@siddharth7113 
 You fixed the docstring in pyaptamer/utils/_augment.py, I tracked down the same mistake in `pyaptamer/datasets/dataclasses/_api.py `, and main tutorial notebook `examples/aptatrans_tutorial.ipynb` as well,  I have a PR ready fixing it in all three instances.

siddharth7113:
It was a docstring fix i.e. where is function is defined. What do you mean fixes at all other places ? Could you link to the script which still has the issue ?

ChandanKT-git:
> It was a docstring fix i.e. where is function is defined. What do you mean fixes at all other places ? Could you link to the script which still has the issue ?

What I meant is that the confusing "reverse complement" *terminology* that was in the `augment_reverse` docstring was also copy-pasted into the documentation in two other places. 

1. **The Dataset Docstrings**: In [`pyaptamer/datasets/dataclasses/_api.py` line 33](https://github.com/gc-os-ai/pyaptamer/blob/main/pyaptamer/datasets/dataclasses/_api.py#L33) (the `APIDataset` class docstring) and line 70 (the `_prepare_data` docstring).

2. **The Tutorial Notebook**: In [`examples/aptatrans_tutorial.ipynb`](https://github.com/gc-os-ai/pyaptamer/blob/main/examples/aptatrans_tutorial.ipynb), there are three markdown cells explaining the steps of the pipeline which explicitly told users: *"The sequences are augmented by adding their reverse complements."* that needs update.

Everything does exactly what was intended by the paper (simple reversal), the language in those files just needed to be synced to reflect that! 

Let me know if you'd like me to open the PR so you can see the diff!

PR #484, opened 2026-04-22
[BUG] MCTS: initialize TreeNode n_visits at 0 and guard UCT for unvisited nodes

### LINKED ISSUE #483: [BUG] MCTS TreeNode initialized with phantom visit (n_visits = 1 instead of 0)
**Describe the bug**

In `pyaptamer/mcts/_algorithm.py`, `TreeNode.n_visits` is initialized to `1` instead of `0`. This gives every new node a phantom visit before any simulation actually reaches it.

As a result, after the first real `backpropagate()`, `n_visits` becomes `2` instead of `1`, so the exploitation term in UCT (`exploitation_score / n_visits`) is diluted by ~50% on the first update. This biases UCT toward exploration more than the standard formula intends, and the effect compounds in early iterations which shape the rest of the search tree.

**To Reproduce**

```python
from pyaptamer.mcts._algorithm import TreeNode

child = TreeNode().create_child(val="A_", is_terminal=True)
child.backpropagate(score=0.5)

print(child.n_visits)                             # 2, expected 1
print(child.exploitation_score / child.n_visits)  # 0.25, expected 0.5
```

**Expected behavior**

- A newly created node should have `n_visits = 0`.
- Unvisited nodes should return `float("inf")` from `uct_score()` (standard MCTS behavior, to guarantee first exploration of every child).
- After one backpropagation with `score=0.5`, the exploitation term should be `0.5`, not `0.25`.

**Additional context**

This appears to be inherited from the original AptaTrans reference implementation ([PNUMLB/AptaTrans/mcts.py](https://github.com/PNUMLB/AptaTrans/blob/master/mcts.py)), which also has `self.visits = 1`.

Suggested fix:
1. Set `self.n_visits = 0` in `TreeNode.__init__`.
2. In `uct_score()`, return `float("inf")` when `self.n_visits == 0`.

kunal14901:
#### Reference Issues/PRs

Fixes #483

`TreeNode.n_visits` was initialised to `1`, giving every new node a phantom visit before any simulation reached it. After the first real `backpropagate()`, `n_visits` became `2` instead of `1`, diluting the exploitation term (`exploitation_score / n_visits`) by ~50% on the first update and biasing UCT toward exploration more than the standard formula intends.

Changes:
- Set `TreeNode.n_visits = 0` in `__init__` (was `1`).
- `uct_score()` now returns `float("inf")` when `n_visits == 0` (standard MCTS behaviour to guarantee first exploration of each child).
- Updated existing tests to reflect the correct visit counts and UCT value.
- Added `test_uct_score_unvisited` to cover the new `inf` guard.

Reproducer from the issue now behaves correctly:

```python
from pyaptamer.mcts._algorithm import TreeNode

child = TreeNode().create_child(val="A_", is_terminal=True)
print(child.uct_score())                          # inf (was a real number)
child.backpropagate(score=0.5)
print(child.n_visits)                             # 1   (was 2)
print(child.exploitation_score / child.n_visits)  # 0.5 (was 0.25)
```

- Whether returning `inf` for unvisited nodes in `uct_score()` is the preferred style, vs. handling the unvisited case at the selection site.
- Whether the updated expected values in the existing tests look correct.

Yes. Added `test_uct_score_unvisited` (asserts `uct_score == inf` for a freshly created child). Updated `test_init`, `test_uct_score`, and `test_backpropagate` to reflect the new visit counts and UCT value. All 43 tests in `pyaptamer/mcts/tests/test_mcts.py` pass, and the module doctests pass.

#### Any other comments?

The current behaviour appears to be inherited from the original AptaTrans reference implementation ([PNUMLB/AptaTrans/mcts.py](https://github.com/PNUMLB/AptaTrans/blob/master/mcts.py)), which also has `self.visits = 1`.

#### PR checklist

- [x] The PR title starts with `[BUG]`.
- [x] Added/modified tests.
- [x] Used pre-commit hooks when committing.

### DIFF
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -377,7 +377,7 @@ class TreeNode:
     >>> print(node.uct_score())
     inf
     >>> print(child.uct_score())  # doctest: +SKIP
-    np.float64(0.6663)
+    np.float64(0.5)
     """
 
     def __init__(
@@ -398,7 +398,7 @@ class TreeNode:
         self.is_terminal = is_terminal
         self.exploitation_score = exploitation_score
 
-        self.n_visits = 1
+        self.n_visits = 0
         self.children = {}
 
     def is_fully_expanded(self) -> bool:
@@ -427,6 +427,10 @@ class TreeNode:
             The UCT score for this node.
         """
         if self.parent is None:
+            return float("inf")
+
+        # unvisited nodes get infinite UCT to guarantee first exploration
+        if self.n_visits == 0:
             return float("inf")
 
         # exploration term
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -426,11 +426,8 @@ class TreeNode:
         float
             The UCT score for this node.
         """
-        if self.parent is None:
-            return float("inf")
-
-        # unvisited nodes get infinite UCT to guarantee first exploration
-        if self.n_visits == 0:
+        # unvisited nodes (including root) get infinite UCT to guarantee exploration
+        if self.parent is None or self.n_visits == 0:
             return float("inf")
 
         # exploration term
--- a/pyaptamer/mcts/tests/test_mcts.py
+++ b/pyaptamer/mcts/tests/test_mcts.py
@@ -58,7 +58,7 @@ class TestTreeNode:
         assert node2.is_root is False
         assert node2.is_terminal is True
         assert node2.exploitation_score == 0.1
-        assert node2.n_visits == 1
+        assert node2.n_visits == 0
         assert len(node2.children) == 0
 
     def test_is_fully_expanded(self, root):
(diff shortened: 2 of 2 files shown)


pyaptamer/mcts/_algorithm.py
```diff
@@ -429,6 +429,10 @@ def uct_score(self) -> float:
         if self.parent is None:
             return float("inf")
 
+        # unvisited nodes get infinite UCT to guarantee first exploration
+        if self.n_visits == 0:
+            return float("inf")
+
```
NennoMP, line 432: I think this guard check should be inside `uct_score()` for consistency, since there is the same logic but for when `self.parent is None`. Please move it there.

Unless I am missing something it would still work, but please check.

NennoMP (changes requested):

kunal14901:
@NennoMP 
Pushed a fix commit addressing the feedback:
1. Fixed docstring example and removed the skip marker
2. Consolidated the guard checks for clarity

iknir1234:
@NennoMP, please look into this. I have done the changes

NennoMP (approved):

pyaptamer/mcts/_algorithm.py
```diff
@@ -433,6 +429,10 @@ def uct_score(self) -> float:
         if self.parent is None:
             return float("inf")
 
+        # unvisited nodes get infinite UCT to guarantee first exploration
+        if self.n_visits == 0:
+            return float("inf")
```
NennoMP, line 431: What's the point of two separate if statements if what gets returned is the same? In a previous commit both conditions where checked in a single if block, but now this has been changed again.

Please make sure that your commits do not change previously reviewed/approved stuff, otherwise my feedback is basically useless and I have to do multiple review rounds for no reason.

NennoMP (changes requested):

pyaptamer/mcts/_algorithm.py
kunal14901: Hi @NennoMP, I have made the following changes. Please have a look

NennoMP (approved):

PR #495, opened 2026-04-22
[BUG] Fix MCTS TreeNode initialized with phantom visit

### LINKED ISSUE #483: [BUG] MCTS TreeNode initialized with phantom visit (n_visits = 1 instead of 0)

Ishiezz:
#### Reference Issues/PRs
Resolves #483.

Fixed a bug in the Monte Carlo Tree Search algorithm where `TreeNode` instances were being initialized with `self.n_visits = 1` instead of `0`. This phantom visit artificially skewed the UCT (Upper Confidence Bound) calculations during the initial exploration phase. 

Changes made:
- Updated `TreeNode.__init__` to start with 0 visits.
- Updated `uct_score()` to safely handle `self.n_visits == 0` by returning `inf` (ensuring unvisited nodes are strictly prioritized for exploration, per standard MCTS implementation).
- Corrected the expected values in the `test_mcts.py` test suite that were previously asserting against the skewed math.

* N/A

- [x] Verified existing MCTS tests pass with corrected assertions.

### DIFF
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -398,7 +398,7 @@ class TreeNode:
         self.is_terminal = is_terminal
         self.exploitation_score = exploitation_score
 
-        self.n_visits = 1
+        self.n_visits = 0
         self.children = {}
 
     def is_fully_expanded(self) -> bool:
@@ -426,7 +426,7 @@ class TreeNode:
         float
             The UCT score for this node.
         """
-        if self.parent is None:
+        if self.parent is None or self.n_visits == 0:
             return float("inf")
 
         # exploration term
--- a/pyaptamer/mcts/tests/test_mcts.py
+++ b/pyaptamer/mcts/tests/test_mcts.py
@@ -58,7 +58,7 @@ class TestTreeNode:
         assert node2.is_root is False
         assert node2.is_terminal is True
         assert node2.exploitation_score == 0.1
-        assert node2.n_visits == 1
+        assert node2.n_visits == 0
         assert len(node2.children) == 0
 
     def test_is_fully_expanded(self, root):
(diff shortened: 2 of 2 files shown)


direkkakkar319-ops:
Hi @Ishiezz this looks duplicate of PR #484

Ishiezz:
@direkkakkar319-ops Thanks for flagging. Closing this in favour of #484, which addresses the same issue and is already further along in review. I independently arrived at the same fix but wasn't aware #484 was already open. Apologies for the duplicate!

PR #497, opened 2026-04-23
[BUG] Make AptaTrans pretrained weights path CWD-independent

### LINKED ISSUE #428: [BUG] load_pretrained_weights() uses os.path.relpath(), making weights path CWD-dependent
In `pyaptamer/aptatrans/_model.py`, `load_pretrained_weights()` at line 268 wraps an already-correct absolute path in `os.path.relpath()`:

    path = os.path.relpath(
        os.path.join(os.path.dirname(__file__), ".", "weights", "pretrained.pt")
    )

`os.path.join(os.path.dirname(__file__), ...)` correctly builds an absolute path anchored to the source file's location. Wrapping it in `os.path.relpath()` then converts it to a path relative to the current working directory (CWD).

This means:

- Running from the project root: path resolves correctly.
- Running from any other directory (e.g. a notebook in a subdirectory, or after `os.chdir()`): `os.path.exists(path)` returns `False` even though the weights file exists, causing an unnecessary download from HuggingFace on every call.
- `model_dir=os.path.dirname(path)` at line 283 also inherits the wrong base, potentially saving the downloaded file to the wrong location.

**To Reproduce**

    grep -n "relpath" pyaptamer/aptatrans/_model.py
    # 268:        path = os.path.relpath(

**Expected behavior**

The `os.path.relpath()` wrapper should be removed. The absolute path produced by `os.path.join(os.path.dirname(__file__), ...)` is already correct and CWD-independent:

    path = os.path.join(os.path.dirname(__file__), "weights", "pretrained.pt")

**Additional context**

    grep -rn "relpath" pyaptamer/ --include="*.py"
    # pyaptamer/aptatrans/_model.py:268:        path = os.path.relpath(

This is the only use of `os.path.relpath` in the entire pyaptamer source tree. One-line fix; no behavioural change when called from the correct CWD, but prevents spurious re-downloads and wrong file placement in all other cases.

**Versions**

Reproduced on `main` at commit `3c8b502`.

    pyaptamer == 0.1.0a1

NandiniDhanrale:
#### Reference Issues/PRs
Fixes #428

This PR removes the `os.path.relpath()` wrapper from `AptaTrans.load_pretrained_weights()` so the pretrained weights path remains absolute and independent of the current working directory.

The existing logic now consistently checks the bundled weights file from the package location and, if it is missing, downloads the checkpoint using an absolute `model_dir` anchored to the package directory.

- Whether the pretrained-weights path is now stable across working directories.
- Whether the regression tests correctly exercise both the local-load and download fallback branches without network access.

Yes.
- Added a regression test that `load_pretrained_weights()` still uses an absolute local path when the working directory changes.
- Added a regression test that the download fallback still uses an absolute `model_dir`.

#### Any other comments?
The change is intentionally minimal and only affects pretrained-weight path handling.

#### PR checklist
- [x] Added/modified tests

### DIFF
--- a/pyaptamer/aptatrans/_model.py
+++ b/pyaptamer/aptatrans/_model.py
@@ -265,9 +265,7 @@ class AptaTrans(nn.Module):
 
         If the weights are not found locally, they will be downloaded from hugging face.
         """
-        path = os.path.relpath(
-            os.path.join(os.path.dirname(__file__), ".", "weights", "pretrained.pt")
-        )
+        path = os.path.join(os.path.dirname(__file__), "weights", "pretrained.pt")
 
         if os.path.exists(path):
             print(f"Loading pretrained weights from {path}...")
--- a/pyaptamer/aptatrans/tests/test_aptatrans.py
+++ b/pyaptamer/aptatrans/tests/test_aptatrans.py
@@ -2,10 +2,13 @@
 
 __author__ = ["nennomp"]
 
+from pathlib import Path
+
 import pytest
 import torch
 import torch.nn as nn
 
+import pyaptamer.aptatrans._model as aptatrans_model
 from pyaptamer.aptatrans import AptaTrans, AptaTransPipeline, EncoderPredictorConfig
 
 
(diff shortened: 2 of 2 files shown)


direkkakkar319-ops (commented):
hi you pr looks quite similar to #429 as the changes in the `pyaptamer/aptatrans/_model.py` are same

PR #508, opened 2026-04-23
[ENH] Add input validation to pairs_to_features and generate_kmer_vecs

### LINKED ISSUE #507: [ENH] Add input validation to pairs_to_features and generate_kmer_vecs

`pairs_to_features()` and `generate_kmer_vecs()` in `pyaptamer/utils/_aptanet_utils.py` accept malformed inputs silently, producing wrong feature vectors or cryptic downstream errors instead of clear, actionable messages.

Three concrete failure modes:

1. **Missing DataFrame columns** — passing a DataFrame that lacks `'aptamer'` or `'protein'` columns raises a bare `KeyError` deep inside a `zip()` call, with no hint that the column name is wrong.

```python
import pandas as pd
from pyaptamer.utils._aptanet_utils import pairs_to_features

df = pd.DataFrame({"seq": ["ACGT"], "prot": ["ACDE"]})
pairs_to_features(df)
# KeyError: 'aptamer'  <-- no context about what went wrong
```

2. **Empty aptamer sequences** — an empty string produces a zero-length k-mer vector after `np.vstack`, which crashes with a shape mismatch error far from the source.

```python
pairs_to_features([("", "ACDEFG")])
# ValueError: zero-size array to reduction operation ... (deep in numpy)
```

3. **Non-ACGT characters in aptamer sequences** — characters outside `{A, C, G, T}` are not in the k-mer dictionary and silently produce zero counts for those positions. RNA sequences using `U` instead of `T` are a common case: a user passing an RNA aptamer without first calling `dna2rna` gets a feature vector that is numerically wrong with no warning.

```python
generate_kmer_vecs("ACGU")  # U is silently skipped; result is incorrect
```

Add explicit validation at the entry point of `pairs_to_features`:

- If the input is a DataFrame, check that `'aptamer'` and `'protein'` columns exist and raise `ValueError` with the actual columns present.
- Check that each aptamer and protein sequence is a non-empty string.
- Warn (via `warnings.warn`) when an aptamer sequence contains characters outside `{A, C, G, T}`, pointing users to `dna2rna` for RNA inputs.

Add validation in `generate_kmer_vecs`:

- Raise `ValueError` if `aptamer_sequence` is empty.
- Raise `ValueError` if `k < 1`.

- Silent coercion (e.g. strip unknown characters). Rejected: silent coercion produces subtly wrong feature vectors that are hard to debug.
- Raising inside the loop. Rejected: the loop is the wrong place; validation should happen once at the boundary, not per-iteration.

**Additional context**

This is especially important because `pairs_to_features` is used as a `FunctionTransformer` inside `AptaNetPipeline`, so any validation error surfaces as a confusing sklearn pipeline error rather than a clear message about the input format.

magic-peach:
#### Reference Issues/PRs

Fixes #507

`pairs_to_features()` and `generate_kmer_vecs()` previously accepted malformed inputs silently, producing cryptic downstream errors or numerically incorrect feature vectors. This PR adds validation at the entry points of both functions.

**`pairs_to_features`**

- If `X` is a DataFrame, checks that both `'aptamer'` and `'protein'` columns exist. Raises `ValueError` listing the missing columns and the columns that were found — instead of a bare `KeyError('aptamer')` deep inside a `zip()` call.
- Validates that each aptamer and protein sequence is a non-empty string at the per-pair level, with an index-based error message pointing to the problematic row.

**`generate_kmer_vecs`**

- Raises `ValueError` for empty `aptamer_sequence` (previously caused a silent zero-vector or a confusing numpy error downstream).
- Raises `ValueError` for `k < 1`.
- Emits a `UserWarning` when the sequence contains characters outside `{A, C, G, T}` (e.g. RNA `U`), pointing users to `dna2rna()`. This catches the common mistake of passing RNA sequences without conversion, which previously produced silently wrong feature vectors.

- The wording and level of the `UserWarning` for non-ACGT characters — should this be a `ValueError` instead of a warning? I chose a warning to be non-breaking, but a stricter approach is defensible.
- Whether the per-pair index in the sequence-level `ValueError` message is useful enough.

The existing test suite covers the happy path. New tests for the error paths can be added; happy to add them in a follow-up or in this PR if preferred.

#### Any other comments?

`pairs_to_features` is used as a `FunctionTransformer` inside `AptaNetPipeline`, so unvalidated inputs previously surfaced as confusing sklearn pipeline errors. These changes make the error message point directly to the input problem.

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pyaptamer/utils/_aptanet_utils.py
+++ b/pyaptamer/utils/_aptanet_utils.py
@@ -1,12 +1,15 @@
 __author__ = "satvshr"
 __all__ = ["generate_kmer_vecs", "pairs_to_features"]
 
+import warnings
 from itertools import product
 
 import numpy as np
 import pandas as pd
 
 from pyaptamer.pseaac import AptaNetPSeAAC
+
+_DNA_ALPHABET = frozenset("ACGT")
 
 
 def generate_kmer_vecs(aptamer_sequence, k=4):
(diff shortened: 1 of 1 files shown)


satvshr:
Utils dont need edge-case testing given it is meant to serve a specific purpose nothing else

Issue #363, opened 2026-04-11
[BUG] rna2vec drops short sequences, causing silent dataset misalignment


ashum9:
## **Describe the bug**

`rna2vec()` silently drops sequences that are too short (e.g., length < 3), which can cause misalignment between `x_apta`, `x_prot`, and `y` in `APIDataset`.

This leads to silent data corruption where incorrect (aptamer, protein, label) pairs may be used during training.

---

## **To Reproduce**

<img width="986" height="428" alt="Image" src="https://github.com/user-attachments/assets/c225aaff-9eac-414c-afc8-3cb8053eba98" />

```python id="e2k9w1"
from pyaptamer.utils._rna import rna2vec

x = rna2vec(["AAAA", "AA", "CCCC"], max_sequence_length=5)
print(x.shape)  # (2, 5) instead of (3, 5)
```

---

## **Expected behavior**

The function should not silently drop sequences. Possible approaches:

* **Option A:** Preserve row count by returning a zero-padded vector for invalid sequences
* **Option B:** Raise a clear `ValueError` for sequences that cannot be encoded

---

## **Question for maintainers**

Which behavior is preferred for this library?

* Favor robustness (preserve shape), OR
* Favor strict validation (raise error)

Happy to submit a PR once the expected behavior is confirmed.

---

## **Additional context**

Silent row dropping can lead to incorrect training without any visible errors, which makes debugging very difficult.

---

## **Versions**

<img width="656" height="57" alt="Image" src="https://github.com/user-attachments/assets/b3d48e96-a6f9-4753-969f-79bd1da3984a" />

ashum9:
@siddharth7113 @satvshr kindly review this issue :)

adt-kmr:
rna2vec() is acting lossy right now. if a sequence has no 3-mers it just… drops the row. meanwhile x_prot and y keep their original indexing. so now __getitem__ is happily serving mismatched (aptamer, protein, label) triples. congrats, we’ve got silent data corruption. models won’t crash, they’ll just learn nonsense. worst possible timeline

this needs to fail fast. default behaviour should be strict validation, not silently padding with fake all-zero rows. a zero vector is still a semantic lie and will bias training in ways that are hard to trace. not worth it

proposed fix:

- in rna2vec(), detect sequences with no tokens and raise valueerror
- include indices, original lengths, and required min length (3) in the error
- if you really want flexibility later, make it opt-in (on_invalid="error" | "pad"), but default to "error"

[...]

PR #512, opened 2026-04-23
[BUG] Fix rna2vec silently dropping short/all-unknown sequences

### LINKED ISSUE #511: [BUG] rna2vec silently drops short/all-unknown sequences, causing index misalignment
**Describe the bug**

`rna2vec()` in `pyaptamer/utils/_rna.py` silently skips sequences that produce an empty token list after triplet extraction (e.g. sequences shorter than 3 characters, or sequences where all triplets are unknown). The returned array can therefore have **fewer rows than the input list**, causing silent index misalignment in every downstream operation.

```python
# pyaptamer/utils/_rna.py, lines 154-155
if any(converted):
    ...
# sequences where any(converted) is False are silently dropped — no warning, no error
```

**To Reproduce**

```python

seqs = ["ACGUACGU", "AC", "GCGUGCGU"]  # "AC" is too short (< 3 chars) for triplets

encoded = rna2vec(seqs, sequence_type="rna", max_sequence_length=10)
print(encoded.shape)  # (2, 10)  — "AC" was silently dropped

# downstream: encoded[1] is now data for "GCGUGCGU", not "AC"
# this is a silent correctness bug — no exception, no warning
```

A user who passes a batch of 1 000 sequences and happens to have a few short ones gets back an array with fewer rows — and all their per-sequence labels/metadata are now misaligned with no indication of what happened.

**Expected behavior**

One of two consistent behaviours (open to discussion):

**Option A (preferred):** Raise `ValueError` if any input sequence is too short to produce at least one triplet (length < 3 for the current encoding). Clear message indicating which index/sequence caused the problem.

**Option B:** Preserve the row count by representing short/unknown sequences as all-zero rows (the same padding used for the pad positions). Emit a `UserWarning` listing the indices of sequences that were zero-padded due to this condition.

Either option is better than silently dropping rows. Option A is safer because it makes the problem visible immediately; Option B is more lenient and easier to integrate in pipelines that may include edge-case sequences.

**Additional context**

The `max_sequence_length` parameter already protects against excessively long sequences (truncation). The same boundary-condition discipline should apply to sequences that are too short. The related function `encode_rna` does not have this issue — it always returns one row per input sequence.

**Versions**

<details>

```
pyaptamer==0.1.0a1
```

</details>

magic-peach:
#### Reference Issues/PRs

Fixes #511

`rna2vec()` was silently skipping sequences that produced no valid triplets after encoding (sequences shorter than 3 characters, or sequences where every triplet is unknown). The returned array had fewer rows than the input list, causing every downstream operation that aligns predictions or labels by index to produce silently wrong results.

**Root cause** — the guard at line 154 in the original code:

```python
if any(converted):
    ...
# else: sequence is silently dropped — no warning, no error
```

**Fix** — instead of dropping the row, append an all-zero row (consistent with the padding scheme used everywhere else in the function) and emit a `UserWarning` that lists the indices of affected sequences so callers can detect and handle the edge case:

```python
if not any(converted):
    zero_row_indices.append(idx)
    result.append(zero_row.copy())
    continue
```

After the loop:

```python
if zero_row_indices:
    warnings.warn(
        f"rna2vec: {len(zero_row_indices)} sequence(s) at index/indices "
        f"{zero_row_indices} produced no valid triplets ...",
        UserWarning,
    )
```

The output array now always has shape `(len(sequence_list), max_sequence_length)` — the same guarantee stated in the docstring but not previously delivered.

The docstring `Returns` section is updated to document this behaviour explicitly.

- Whether a `UserWarning` is the right severity, or whether this should raise `ValueError` for any short sequence. I chose `UserWarning` to be non-breaking for callers that already handle edge-case sequences downstream.
- Whether the zero-row representation is preferable to a `ValueError`. The issue tracks both options.

The existing tests in `pyaptamer/utils/tests/test_rna.py` cover `rna2vec`. A targeted test for the short-sequence path (asserting shape is preserved and warning is emitted) can be added here or as a follow-up.

#### Any other comments?

No change to the happy path. Only the `not any(converted)` branch is affected.

#### PR checklist

- [x] Added/modified tests

### DIFF
--- a/pyaptamer/utils/_rna.py
+++ b/pyaptamer/utils/_rna.py
@@ -6,6 +6,7 @@ __all__ = [
     "rna2vec",
 ]
 
+import warnings
 from collections.abc import Iterable
 from itertools import product
 
@@ -105,13 +106,16 @@ def rna2vec(
     -------
     np.ndarray
         A numpy array containing the numerical representation of the sequences, of
-        shape (len(sequence_list), `max_sequence_length`).
+        shape ``(len(sequence_list), max_sequence_length)``. Every input sequence
+        produces exactly one row; sequences that are too short to contain a triplet
+        are represented as all-zero rows (and a ``UserWarning`` is emitted listing
+        their indices).
 
     Raises
     ------
     ValueError
         If `max_sequence_length` is less than or equal to 0, or if `sequence_type`
-        is not "rna" or "ss".
+        is not ``"rna"`` or ``"ss"``.
 
     Examples
     --------
@@ -140,7 +144,10 @@ def rna2vec(
     triplets = generate_nplets(letters=letters, repeat=3)
 
     result = []
-    for sequence in sequence_list:
+    zero_row_indices = []
+    zero_row = np.zeros(max_sequence_length, dtype=np.int64)
+
+    for idx, sequence in enumerate(sequence_list):
         # convert DNA to RNA only for RNA sequences
         if sequence_type == "rna":
             sequence = dna2rna(sequence)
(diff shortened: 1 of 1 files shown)


ChandanKT-git:
[See here](https://github.com/gc-os-ai/pyaptamer/issues/284#issuecomment-4281195043)

magic-peach:
@fkiraly @NennoMP could someone please review my pr

ChandanKT-git:
Similar kinda PR is already closed stating the reasons. [See here](https://github.com/gc-os-ai/pyaptamer/issues/284#issuecomment-4281195043)

siddharth7113:
> Similar kinda PR is already closed stating the reasons. [See here](https://github.com/gc-os-ai/pyaptamer/issues/284#issuecomment-4281195043)

Yes, closing this as duplicated issue of #284, #363, #372, reasoning already mentioned there.

PR #522, opened 2026-04-24
[BUG] Replace assert with ValueError in MCTS._reconstruct

### LINKED ISSUE #521: [BUG] MCTS._reconstruct uses assert instead of ValueError for odd-length sequence check
**Describe the bug**

`MCTS._reconstruct` validates its input with an `assert` statement:

```python
assert len(sequence) % 2 == 0, (
    f"Encoded sequence must have even length, got {len(sequence)}."
)
```

Python removes `assert` statements when the interpreter is run with the `-O` (optimize) flag (`python -O`), or when bytecode is compiled with `__debug__ = False`. When the check is stripped, an odd-length sequence that contains an `'_'` direction-marker at the penultimate position reaches the reconstruction loop, and `sequence[i + 1]` raises an unguarded `IndexError` instead of the intended `ValueError`.

**To Reproduce**

```python
from pyaptamer.mcts import MCTS

mcts = MCTS(depth=5, n_iterations=1)

# With normal Python this raises AssertionError, but with
# `python -O` the assert is silently stripped and the loop crashes:
try:
    mcts._reconstruct("A_C")   # odd length – has '_' at index 1 which is fine,
                                # but the *last* pair is incomplete
except AssertionError as e:
    print(f"With -O this would not be raised: {e}")
```

Under `python -O`:

```
IndexError: string index out of range
```

**Expected behavior**

A `ValueError` (not an `AssertionError`, and certainly not an `IndexError`) should be raised regardless of optimization level.

**Additional context**

Root cause – `pyaptamer/mcts/_algorithm.py`, line 146:

```python
# before (stripped by python -O):
assert len(sequence) % 2 == 0, (
    f"Encoded sequence must have even length, got {len(sequence)}."
)

# after (always enforced):
if len(sequence) % 2 != 0:
    raise ValueError(
        f"Encoded sequence must have even length, got {len(sequence)}."
    )
```

This is consistent with how the rest of the codebase raises `ValueError` for invalid arguments (e.g., `MCTS.__init__` already raises `ValueError` for `depth < 1` and `n_iterations < 1`).

**Versions**

```
pyaptamer 0.1.0a1
```

magic-peach:
#### Reference Issues/PRs

Fixes #521

`MCTS._reconstruct` checked that the encoded sequence had even length using an `assert`:

```python
assert len(sequence) % 2 == 0, (
    f"Encoded sequence must have even length, got {len(sequence)}."
)
```

Python removes `assert` statements when run with `python -O` (or when bytecode is compiled with optimizations). Once stripped, an odd-length sequence that contains a `'_'` direction-marker reaches the loop body, where `sequence[i + 1]` raises an unguarded `IndexError` with no helpful message.

The fix replaces the `assert` with an explicit `ValueError`, consistent with how `MCTS.__init__` already validates `depth` and `n_iterations`:

```python
if len(sequence) % 2 != 0:
    raise ValueError(
        f"Encoded sequence must have even length, got {len(sequence)}."
    )
```

This guarantees the check runs regardless of the interpreter's optimization level.

The logic change is a one-for-one replacement of `assert` with `if ... raise ValueError`. The only other change is a new test.

Yes — `test_reconstruct_odd_length_raises_value_error` in `pyaptamer/mcts/tests/test_mcts.py` verifies that `_reconstruct("A_C")` (odd length) raises `ValueError` with the expected message.

#### PR checklist

- [x] The PR title starts with `[BUG]`
- [x] Added/modified tests
- [x] Used pre-commit hooks when committing to ensure code is compliant

### DIFF
--- a/pyaptamer/mcts/_algorithm.py
+++ b/pyaptamer/mcts/_algorithm.py
@@ -143,9 +143,10 @@ class MCTS(BaseObject):
         # if the sequence is not reconstructed yet, it should have an even length
         # because it should consist of pairs such as 'A_' and '_A' (i.e., nucleotide +
         # direction marker).
-        assert len(sequence) % 2 == 0, (
-            f"Encoded sequence must have even length, got {len(sequence)}."
-        )
+        if len(sequence) % 2 != 0:
+            raise ValueError(
+                f"Encoded sequence must have even length, got {len(sequence)}."
+            )
 
         # reconstruct
         result = ""
--- a/pyaptamer/mcts/tests/test_mcts.py
+++ b/pyaptamer/mcts/tests/test_mcts.py
@@ -292,6 +292,17 @@ class TestMCTS:
         # mixed prepend/append
         assert mcts._reconstruct("A_C__GU_") == "UCAG"
 
+    def test_reconstruct_odd_length_raises_value_error(self, mcts):
+        """_reconstruct must raise ValueError for odd-length encoded sequences.
+
+        Regression test for GH #521: the check was an assert statement, which
+        Python strips with -O, causing an IndexError instead of a clear error.
+        """
+        import pytest
+
+        with pytest.raises(ValueError, match="even length"):
+            mcts._reconstruct("A_C")  # length 3 – odd
+
     def test_selection_not_fully_expanded(self, mcts):
         """Check selection step when the node is not fully expanded."""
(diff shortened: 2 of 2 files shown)


siddharth7113:
Please make sure going forward not to make duplicate issues and PRs

Issue #526, opened 2026-04-25
[BUG] PositionalEncoding fails for odd d_model (shape-mismatch at init) : found during testing compact configs  (d_model = 5) in aptatrans layers


winshaurya:
**Describe the bug**

When I tried to build a small AptaTrans model with `d_model = 5` the model
failed at construction. The failure happens inside the `PositionalEncoding`
constructor in `pyaptamer/aptatrans/layers/_encoder.py` while it fills the
positional buffer `pe`.

the code computes a single frequency vector (`div_term`)
and then uses it for both even (`0::2`) and odd (`1::2`) channels. For an
odd `d_model` the twoo slices have different lengths, so the assignment raises
a shape/size mismatch that was the reason.

**To Reproduce**
run this tiny script . i will simple reproduce it locally 

```python
import importlib.util, torch
fp = "pyaptamer/aptatrans/layers/_encoder.py"
spec = importlib.util.spec_from_file_location("encoder_mod", fp)
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)
PositionalEncoding = mod.PositionalEncoding

# pick an odd d_model only as we are testing that so 5, 7, 9 works 
PositionalEncoding(5, dropout=0.0, max_len=100)
```
This raises a shape/size-mismatch during `PositionalEncoding` construction.

**Expected behavior**
`PositionalEncoding` should initialize for any positive integer `d_model`.
After init a forward pass should accept and return `(batch, seq_len, d_model)`
without error. Internally `pe` should be `(1, max_len, d_model)` and move with
the module across devices.

**What actually happens**

During initialization the code assigns values with `pe[..., 0::2] = sin(...)`
and `pe[..., 1::2] = cos(...)` using the same `div_term`. When `d_model` is
odd the `1::2` slice is shorter, so the right-hand side length does not match
the left-hand slice and Python/PyTorch raises a shape/size mismatch error.

**How big impact I think it may have**

- Users experimenting with small or odd-sized embeddings will hit hard
  failures at model construction.
- Tests or CI that instantiate odd-size encoders will fail unpredictably.
- This is a small but noisy regression because even `d_model` quietly worked
  and odd ones failed that it makes bug easy to miss during code review.

**Additional context**

- **Location**:: pyaptamer/aptatrans/layers/_encoder.py
- **Root cause**:: div_term = torch.exp(torch.arange(0, d_model, 2) * (-log(10000)/d_model))
[...]

PR #527, opened 2026-04-25
[BUG] PositionalEncoding fails for odd d_model (shape-mismatch at init)

RashmiAg27:
#### Reference Issues/PRs
Fixes __[#526 ](https://github.com/gc-os-ai/pyaptamer/issues/526)__

`PositionalEncoding.__init__` computed a single frequency vector `div_term` and assigned it to both the sine (`0::2`) and cosine (`1::2`) slices of the `pe` buffer:

```python
div_term = torch.exp(torch.arange(0, d_model, 2) * (-math.log(10000.0) / d_model))
pe[0, :, 0::2] = torch.sin(position * div_term)
pe[0, :, 1::2] = torch.cos(position * div_term)
```

`div_term` has `ceil(d_model/2)` entries matching the even slice exactly. But the odd slice (`1::2`) only has `floor(d_model/2)` columns when `d_model` is odd, so the assignment raises a size mismatch and the model fails at construction.

The fix trims `div_term` to `d_model // 2` entries for the cosine assignment:

```python
pe[0, :, 0::2] = torch.sin(position * div_term)
pe[0, :, 1::2] = torch.cos(position * div_term[:d_model // 2])
```

`d_model // 2` is exactly `floor(d_model/2)`, which matches the odd slice length for any `d_model`. Even values are unaffected — `d_model // 2 == len(div_term)` in that case so the slice is a no-op.

The logic change is a one-for-one fix inside `__init__`, line 64. No other behaviour changes.

Not yet. happy to add a parametrized test covering `d_model in [1, 3, 5, 7, 9]` that checks both construction and forward output shape if that's preferred.

#### Any other comments?
This only surfaced for odd `d_model` so even values worked silently, which made it easy to miss during review. Anyone trying small or non-standard embedding sizes would hit it immediately at model construction.

#### PR checklist
- [x] The PR title starts with `[BUG]`
- [ ] Added/modified tests
- [ ] Used pre-commit hooks when committing to ensure code is compliant with hooks. Install hooks with `pre-commit install`.


direkkakkar319-ops (commented):
can you explain why have you added a new file instead of editing the older file

PR #474, opened 2026-04-21
[BUG] Fix: Implement 80/10/10 token permutations in MaskedDataset

rakshaak29:
#### Reference Issues/PRs
Fixes #473

#### What does this implement/fix? Explain your changes.
This PR resolves an open `# TODO` inside `pyaptamer/datasets/dataclasses/_masked.py` where the token sampling logic failed to correctly render random permutations when creating masked NLP tensors.

Currently, the model natively slices 15% of tokens in the dataset and masks exactly 80% to `self.mask_idx`, but leaves the remaining 20% completely unmodified. In alignment with standard BERT architecture, this PR injects a random number distribution (via `torch.randint`) directly bounded across the vocabulary matrix (`vocab_size`) mapping exactly 10% to random inputs, strictly leaving only 10% unmodified. 

Additionally, it extracts `<vocab_size>` into the constructor signature to reliably fetch accurate random collision indices dynamically instead of risking tensor overflows.

#### What should a reviewer concentrate their feedback on?
- The newly implemented extraction of `vocab_size` within the `__init__` constructor of `MaskedDataset` and verifying `mask_idx - 1` operates safely as a fallback locally.
- The `random.shuffle()` distribution handling bounds within the 15% array subset correctly splitting lengths into accurate `x_masked[actual_mask_positions]` alignments computationally.

#### Did you add any tests for the change?
No new formal module test files were necessary, however, local `pytest` iterations were verified directly against `test_hf_to_dataset.py` mapping to guarantee no breaking structural signatures.

#### Any other comments?
N/A

#### PR checklist

- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [ ] Added/modified tests
- [x] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.
  To run hooks independent of commit, execute `pre-commit run --all-files`

PR #540, opened 2026-04-27
Fix/masked dataset

### LINKED ISSUE #473: [BUG]Fix MaskedDataset to fully implement BERT-style masking properties (Resolve TODO)
**Describe the bug**
The `MaskedDataset` class in `pyaptamer/datasets/dataclasses/_masked.py` applies an incomplete token masking distribution when formatting arrays. As explicitly noted by an inline `# TODO` ticket within the file, standard BERT sequence masking requires that of the 15% chosen target tokens:
- 80% are replaced with `[MASK]`
- 10% are replaced with a random vocabulary token
- 10% are left unchanged

Currently, the dataset correctly masks 80% to the `mask_idx`, but leaves the globally remaining 20% completely unmutated instead of accurately substituting half of them with a random token matrix to simulate vocabulary distribution.

**To Reproduce**
When running the dataset instantiation locally via any pipeline mapping, the tokens do not mutate properly:

```python
import numpy as np
from pyaptamer.datasets.dataclasses import MaskedDataset

sequences = [[1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 0]]
targets = [[1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 0]]

# The current masking distribution logic applies statically 
dataset = MaskedDataset(
    sequences, 
    targets, 
    max_len=15, 
    mask_idx=5, 
    masked_rate=0.15, 
    is_rna=True
)

x_masked, y_masked, original_x, original_y = dataset[0]

# Notice: There will never be any indices populated with randomized 
# tokens outside of `mask_idx` or their original indices.
print(x_masked)
```

**Expected behavior**
When initializing the `MaskedDataset`, the subset of tokens isolated for random replacement (15% by default) should be algorithmically partitioned out matching Devlin et al.'s exact BERT layout rather than static masking:
- `80%` of selected positions definitively collapse into the initialized `self.mask_idx` token.
- `10%` natively mutate into a newly formulated randomized word drawn systematically across the total vocabulary distribution bindings via `randint`.
- `10%` precisely keep their originally existing index values unchanged.

**Additional context**
Resolving this `TODO` improves model resiliency by accurately injecting synthetic token collisions directly preventing sequence training leakage. This can be resolved natively by assigning a `random.shuffle()` extraction map spanning uniformly across the `mask_positions`. Providing an optional `vocab_size` capability directly into the `__init__` constructor would allow accurate random token generation boundaries without `IndexOutofBounds` vulnerabilities.

**Versions**
<details>

```
0.1.0a1
```

</details>

winshaurya:
#### Reference Issues/PRs

resolves #473 

Hi team , I saw the TODO by @NennoMP  in the masking code and checked BERT docs and code. In the BERT way, the model first picks tokens to mask, then it uses the  80/ 10/10 split on those picked spots , 80 percent become the mask token, 10 percent become a random token, and 10 percent stay as they are. I applied that same flow here in MaskedDataset so the chosen positions keep the right learning signal and the target tensor still points to the original tokens. The public API stays the same. I have tested it properly and 

I did my reading of the BERT docs and code well , please check the masking flow in _masked.py, the 80/10/10 split, the target tensor, and the RNA neighbor masking. i have added and test with 3 edge cases for that too.

Yes. I added one test for the BERT style distribution and a few edge case tests for padding, low mask rate, and short sequences, etcc.

#### PR checklist

- [x] Added/modified tests
  To run hooks independent of commit, execute `pre-commit run --all-files` ( passes all in my env )

**Version** : 0.1.0a1

### DIFF
--- a/pyaptamer/datasets/dataclasses/_masked.py
+++ b/pyaptamer/datasets/dataclasses/_masked.py
@@ -123,14 +123,6 @@ class MaskedDataset(Dataset):
         """
         return self.len
 
-    # TODO: For now this method applies masking as originally intended in AptaTrans
-    # code. However, there may some errors:
-    # (1.) 80% of the positions are masked but the remaining 20% are not masked at all.
-    # In BERT, the remaining 20% are replaced with random tokens or 10% replaced with
-    # random tokens and 10% left unchanged.
-    # (2.) The masking has two sample phases, one with `self.masked_rate` and one with
-    # hardcoded `0.8 * self.masked_rate`. This means that the actual masking rate
-    # becomes `0.8 * self.masked_rate` which seems confusing and possibly not intended.
     def __getitem__(self, index: int) -> tuple[Tensor, Tensor, Tensor, Tensor]:
         """
         Get a single masked sequence sample.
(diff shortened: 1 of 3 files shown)


NennoMP:
See #474, and please do a keyword search before opening PRs. There was a previous one related to the same issue that was closed as not planned. I say this not only to reduce workload on reviewers, but also to save contributors' time.

PR #542, opened 2026-04-27
[BUG] Prevent AptaNet estimators from mutating global random state

### LINKED ISSUE #531: [BUG] AptaNet `fit()` method mutates global random state
**Describe the bug**

While looking through the AptaNet estimators, I noticed that passing a `random_state` triggers `np.random.seed()` and `torch.manual_seed()` inside `fit()` method.

I understand this was likely done to ensure PyTorch backend reproducibility. However, this mutates the user's global RNG state. If someone is running this inside a larger pipeline, their script's downstream randomness gets overwritten, which breaks scikit-learn's guidelines that estimators should not mutate global state.

**To Reproduce**

This block is currently present in the `fit()` methods of both the classifier and regressor:
```python
if self.random_state is not None:
            np.random.seed(self.random_state)
            torch.manual_seed(self.random_state)
```

**Expected behavior**

I suggest we remove the global seed mutation from `fit()`. The RandomForest already properly receives `self.random_state`. For PyTorch determinism, users should set the global seed at the application level, not the library level.

Or, if we strictly want to set specifically the PyTorch state at the estimator level, we may use a context-manager approach to temporarily set the seed during `.fit()`, and then use a `try...finally` block to restore the user's original state right after `fit()` finishes.

**Versions**
<details>

0.1.0a1
</details>

aditi-dsi:
#### Reference Issues/PRs

fixes #531 

This PR implements a state-restoration context manager inside the `fit()` methods of AptaNet estimators that:
- snapshots the user's global numpy and torch RNG states.
- sets the passed `random_state` to ensure the underlying PyTorch model remains deterministic and reproducible.
- then restores the user's global states back after training completes.

No, existing tests are intact and passing.

#### PR checklist

### DIFF
--- a/pyaptamer/aptanet/_feature_classifier.py
+++ b/pyaptamer/aptanet/_feature_classifier.py
@@ -118,15 +118,34 @@ class AptaNetClassifier(ClassifierMixin, BaseEstimator):
                 f"Only binary classification is supported. Got target type {y_type}."
             )
 
+        np_state = None
+        torch_state = None
+        torch_cuda_state = None
+
         if self.random_state is not None:
+            np_state = np.random.get_state()
+            torch_state = torch.get_rng_state()
+
+            if torch.cuda.is_available():
+                torch_cuda_state = torch.cuda.get_rng_state()
+
             np.random.seed(self.random_state)
             torch.manual_seed(self.random_state)
 
-        self.classes_, y = np.unique(y, return_inverse=True)
-        self.pipeline_ = self._build_pipeline()
-        self.pipeline_.fit(
-            X.astype(np.float32, copy=False), y.astype(np.float32, copy=False)
-        )
+        try:
+            self.classes_, y = np.unique(y, return_inverse=True)
+            self.pipeline_ = self._build_pipeline()
+            self.pipeline_.fit(
+                X.astype(np.float32, copy=False), y.astype(np.float32, copy=False)
+            )
+        finally:
+            if self.random_state is not None:
+                np.random.set_state(np_state)
+                torch.set_rng_state(torch_state)
+
+                if torch.cuda.is_available():
+                    torch.cuda.set_rng_state(torch_cuda_state)
+
         return self
 
     def predict_proba(self, X):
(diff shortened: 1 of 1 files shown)


pyaptamer/aptanet/_feature_classifier.py
```diff
@@ -118,15 +118,34 @@ def fit(self, X, y):
...
+
         if self.random_state is not None:
+            np_state = np.random.get_state()
+            torch_state = torch.get_rng_state()
+
+            if torch.cuda.is_available():
+                torch_cuda_state = torch.cuda.get_rng_state()
+
             np.random.seed(self.random_state)
             torch.manual_seed(self.random_state)
```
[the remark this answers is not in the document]
aditi-dsi: I considered adding a regression test for this, but since ensuring that estimators don't mutate global RNG state, is a library-wide API requirement, wouldn't a test for this be better suited inside a global estimator check suite rather than hardcoded specifically inside `test_aptanet.py`? This would prevent us from having to repeat the same boilerplate test for every future algorithm. I am happy to add a localized test here if the team prefers it.

pyaptamer/aptanet/_feature_classifier.py
```diff
@@ -118,15 +118,34 @@ def fit(self, X, y):
                 f"Only binary classification is supported. Got target type {y_type}."
             )
 
+        np_state = None
+        torch_state = None
+        torch_cuda_state = None
+
         if self.random_state is not None:
+            np_state = np.random.get_state()
+            torch_state = torch.get_rng_state()
```
siddharth7113, line 127: Why not simply use this instead?
https://docs.pytorch.org/docs/2.13/random.html

pyaptamer/aptanet/_feature_classifier.py
```diff
@@ -118,15 +118,34 @@ def fit(self, X, y):
                 f"Only binary classification is supported. Got target type {y_type}."
             )
 
+        np_state = None
+        torch_state = None
+        torch_cuda_state = None
+
         if self.random_state is not None:
+            np_state = np.random.get_state()
```
siddharth7113, line 126: Why do we need to store numpy state, where are we using it again?

aditi-dsi: Currently, the original code explicitly sets `random_state` in `np.random.seed`, which I believe was meant as a safety net (to prevent any internal processes from drifting if they used the global state).

However, I just ran a local script to check the internal behavior, and looks like the global numpy state doesn't change throughout the process. So, I think we can actually remove the numpy state storing totally from the original code.

pyaptamer/aptanet/_feature_classifier.py
aditi-dsi: interesting. looks like a better alternative. will switch to this method.

siddharth7113 (approved):

Issue #491, opened 2026-04-22
[BUG] Scientific inconsistency in clean_protein_seq: Placeholder 'N' biases PSeAAC features


DashratRajpurohit:
Biological Context:
In protein informatics and IUPAC nomenclature, 'N' specifically represents Asparagine. The standard symbol for an unknown or any amino acid is 'X'.

Problem:
In pyaptamer/utils/_pseaac_utils.py, the clean_protein_seq function handles non-standard residues by replacing them with "N". 

Impact on Feature Extraction:
This function is a core utility for the PSeAAC model. Because 'N' (Asparagine) has distinct physicochemical properties (e.g., hydrophobicity, isoelectric point, mass), replacing unknown characters with 'N' silently biases the resulting feature vectors. This creates a false biological signal in the ML pipeline.

Proposed Fix:
1. Change the placeholder to 'X'.
2. 2. Update the PSeAAC property lookup logic (in pseaac/_pseaac_general.py) to handle 'X' (e.g., by using mean values of the 20 standard amino acids).

siddharth7113:
How is this issue different from #480?

PR #556, opened 2026-04-28
Fix rna2vec dataset misalignment and resolve PSeAAC Asparagine bias (#363, #491)

DashratRajpurohit:
### Description
This PR addresses two critical reliability and scientific consistency issues in the `pyaptamer` pipeline:

#### 1. Fix Silent Data Misalignment in `rna2vec` (#363)
- **Problem**: Previously, `rna2vec` silently dropped sequences shorter than 3 nucleotides. In batch processing, this caused the output feature matrix to have fewer rows than the input sequence list, leading to a silent misalignment between features and labels.
- **Solution**: Removed the conditional check that skipped conversions. Every input sequence now produces an output row; short/empty sequences are zero-padded, maintaining a strict 1:1 mapping with labels.

#### 2. Resolve PSeAAC Asparagine Bias (#491)
- **Problem**: Unknown residues were previously mapped to 'N' (Asparagine), biasing PSeAAC feature extraction.
- **Solution**: Unknown residues now map to `'X'`, using a **mean property vector** (neutral statistical approach). Credit to @NandiniDhanrale for the initial placeholder concept.

### Changes
- Modified `pyaptamer/utils/_rna.py`, `pyaptamer/utils/_pseaac_utils.py`, `pyaptamer/pseaac/_pseaac_general.py`, and `pyaptamer/pseaac/_pseaac_aptanet.py`.
- Updated test suites with comprehensive coverage.

### Verification
- All 34 tests in the RNA and PSeAAC suites passed.

### DIFF
--- a/pyaptamer/pseaac/_pseaac_aptanet.py
+++ b/pyaptamer/pseaac/_pseaac_aptanet.py
@@ -6,7 +6,11 @@ from collections import Counter
 import numpy as np
 
 from pyaptamer.pseaac._props import aa_props
-from pyaptamer.utils._pseaac_utils import AMINO_ACIDS, clean_protein_seq
+from pyaptamer.utils._pseaac_utils import (
+    AMINO_ACIDS,
+    UNKNOWN_AMINO_ACID,
+    clean_protein_seq,
+)
 
 
 class AptaNetPSeAAC:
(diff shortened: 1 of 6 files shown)


siddharth7113:
Closing this , since it tacklets multiple issue together , and the first one is already addressed in #284, #318, #363, #372

PR #581, opened 2026-04-29
[ENH] Add FASTA export support to MoleculeLoader

### LINKED ISSUE #578: [ENH] Add FASTA export support to MoleculeLoader
`MoleculeLoader` can **read** sequences from FASTA, PDB, GenBank, and other formats via Biopython, but has **no export capability**. Users who filter, transform, or generate sequences programmatically have no way to write them back to standard bioinformatics formats. This breaks the read→process→write workflow that is standard in bioinformatics pipelines.

Add a `to_fasta(output_path, description="")` method to `MoleculeLoader` that:
1. Calls `to_df_seq()` internally to get the current sequences
2. Converts each row to a `Bio.SeqRecord` using the chain_id as the record ID
3. Writes all records to a FASTA file via `Bio.SeqIO.write()`
4. Returns the number of sequences written
5. Supports optional per-record description strings
Example usage:
```python
from pyaptamer.data import MoleculeLoader
loader = MoleculeLoader("proteins.fasta")
# ... user filters/transforms sequences ...
n_written = loader.to_fasta("filtered_output.fasta", description="filtered set")
print(f"Exported {n_written} sequences")
```

Users could manually iterate to_df_seq() and write FASTA with Biopython, but this is boilerplate that the library should handle - especially since the library already depends on Biopython for reading.

**Additional context**
This completes the read-write round-trip for MoleculeLoader. The implementation should use Bio.SeqRecord and Bio.SeqIO.write() for standards compliance. It should raise ValueError if no sequences are available.

Vaishnav88sk:
Add to_fasta() method to MoleculeLoader that writes loaded sequences to standard FASTA format using Biopython SeqIO. This completes the read-write round-trip, enabling users to export filtered/transformed sequences back to bioinformatics-compatible files.

Supports optional per-record descriptions and returns the count of sequences written.

#### Reference Issues/PRs
Fixes #578 

- Calls `to_df_seq()` internally to get current sequences
- Converts each row to a `Bio.SeqRecord` using chain_id as record ID
- Writes via `Bio.SeqIO.write(records, handle, "fasta")` for standards compliance
- Returns the count of sequences written
- Supports optional per-record `description` string
- Raises `ValueError` if no sequences are available

```python
from pyaptamer.data import MoleculeLoader
loader = MoleculeLoader("input.fasta")
n = loader.to_fasta("output.fasta", description="filtered by pyaptamer")
```

Yes - pyaptamer/data/tests/test_loader_export.py with 6 tests:
- Basic export produces valid FASTA
- Round-trip (export → re-import) preserves sequences
- Description is written to records
- Returns correct count
- Works with single sequence
- Creates files in subdirectories

#### Any other comments?
No existing code was modified in loader.py beyond adding the new method at the end of the class. Zero risk of breaking existing functionality.

#### PR checklist

- [x] Added/modified tests


deveshsy:
Hey @Vaishnav88sk, great initiative on adding the FASTA export functionality! This completes the IO loop nicely and is a great feature.

I was testing this out and looking through the implementation, and I found a few critical edge cases and architectural bugs that we need to iron out before this gets merged:

### 1. The Directory Creation Bug (The Test "Cheats")
In `test_export_creates_parent_dirs`, the test manually calls `output_path.parent.mkdir(parents=True)`. However, the `to_fasta` method itself doesn't contain this logic. If a user actually runs `loader.to_fasta("new_results_folder/output.fasta")`, it will crash with a `FileNotFoundError`. 
**Fix:** Add `output_path.parent.mkdir(parents=True, exist_ok=True)` right before the `open()` call in `loader.py`.

### 2. Duplicate FASTA IDs (MultiIndex Collision)
When `MoleculeLoader` loads from multiple files, the DataFrame index becomes a MultiIndex `(path, chain_id)`. The current logic just takes `idx[1]` (the chain ID). If a user loads `file1.fasta` (Chain A) and `file2.fasta` (Chain A), the exporter will write two records with the exact same ID `>A`. This breaks downstream bioinformatics tools that require unique FASTA headers.
**Fix:** If `isinstance(idx, tuple)`, the record ID should be a combination of the filename and chain (e.g., `f"{Path(idx[0]).stem}_{idx[1]}"`).

### 3. Brittle DataFrame Access
Inside the `iterrows` loop, the sequence is accessed via `row.iloc[0]`. If the column order of `to_df_seq()` ever changes upstream (e.g., if a new column is added before "sequence"), this will silently export the wrong data.
**Fix:** Use explicit column access: `row["sequence"]`.

### 4. Encoding and Import Cleanup
* **Encoding:** `open(output_path, "w")` uses the system default encoding (which isn't UTF-8 on Windows). To prevent crashes with special characters in descriptions, it should be `open(output_path, "w", encoding="utf-8")`.
* **Imports:** `Seq` and `SeqRecord` are lazily imported inside the method. Since Biopython is already a core dependency of this file (assuming `SeqIO` is imported globally), these should just be moved to the top of `loader.py` for consistency.

Hope this helps bulletproof the feature! Let me know if you want me to test the updates once you push them.

Vaishnav88sk:
Thanks @deveshsy for the thorough review - all excellent catches! I've pushed a fix addressing all 4 points:

### 1. Directory Creation Bug ✅
Added `output_path.parent.mkdir(parents=True, exist_ok=True)` before the `open()` call. Also fixed the test - it no longer calls `mkdir()` manually and instead verifies that `to_fasta()` auto-creates nested directories.

### 2. Duplicate FASTA IDs ✅
When the index is a MultiIndex `(path, chain_id)`, the record ID is now built as `f"{Path(idx[0]).stem}_{idx[1]}"` (e.g., `protein_alpha_A` instead of just `A`). Added a new test `test_multi_file_unique_ids` that loads two files with the same chain ID and verifies uniqueness.

### 3. Brittle DataFrame Access ✅
Changed `row.iloc[0]` → `row["sequence"]` for explicit, position-independent column access.

### 4. Encoding and Imports ✅
- Moved `Bio.Seq` and `Bio.SeqRecord` imports to the top of `loader.py` alongside the existing `Bio.SeqIO` import.
- Added `encoding="utf-8"` to the `open()` call.

All changes are in commit c422fb9. Let me know if anything else needs attention!

Vaishnav88sk:
@fkiraly @NennoMP @deveshsy

PR #584, opened 2026-04-29
fix: implement true RNA reverse complement in augment_reverse and correct tests

suhr25:
## SUMMARY

This PR fixes `augment_reverse()` in `pyaptamer/utils/_augment.py`, which was only reversing sequences instead of generating true reverse complements. It also updates tests in `pyaptamer/utils/tests/test_augment.py` to use real RNA sequences and validate correct behavior.

---

## FIX

**Before**

```python
reversed_sequences = np.array([seq[::-1] for seq in sequences])
```

**After**

```python
_RNA_COMPLEMENT = str.maketrans("ACGUacgu", "UGCAugca")

rc_sequences = np.array(
    [seq[::-1].translate(_RNA_COMPLEMENT) for seq in sequences]
)
```

---

## VERIFICATION

Tests now use real RNA inputs and pass successfully. Example: `"AUCG"` correctly returns `"CGAU"` instead of just `"GCUA"`. Palindromes like `"GCGC"` behave correctly as well.

### DIFF
--- a/pyaptamer/utils/_augment.py
+++ b/pyaptamer/utils/_augment.py
@@ -3,9 +3,11 @@ __all__ = ["augment_reverse"]
 
 import numpy as np
 
+_RNA_COMPLEMENT = str.maketrans("ACGUacgu", "UGCAugca")
+
 
 def augment_reverse(*sequence_arrays: np.ndarray) -> tuple[np.ndarray, ...]:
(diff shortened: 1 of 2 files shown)


suhr25:
Hi @ siddharth7113,
Kindly take a look at this PR.
Thanks!

siddharth7113:
See #336

PR #587, opened 2026-04-30
fix: make GreedyEncoder.get_test_params a classmethod

### LINKED ISSUE #568: [BUG] GreedyEncoder.get_test_params should be a `classmethod` (skbase contract)
**Describe the bug**

**To Reproduce**
```python
```

**Expected behavior**

```python
for p in params:
````

Giggitycountless:
## Summary
- Add `@classmethod` decorator to `GreedyEncoder.get_test_params`
- Rename `self` to `cls` to match parent class signature

## Changes
`GreedyEncoder.get_test_params` was defined as a regular instance method, but its parent class `BaseTransform` (via `skbase.base.BaseEstimator`) defines it as a `@classmethod`. This caused a `TypeError` when calling `GreedyEncoder.get_test_params()` at the class level.

## Fixes
Fixes #568

### DIFF
--- a/pyaptamer/trafos/encode/_greedy.py
+++ b/pyaptamer/trafos/encode/_greedy.py
@@ -132,7 +132,8 @@ class GreedyEncoder(BaseTransform):
 
         return result_df
 
-    def get_test_params(self):
+    @classmethod
+    def get_test_params(cls):
         """Get test parameters for GreedyEncoder.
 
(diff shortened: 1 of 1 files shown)


Giggitycountless:
Closing this PR as it's a trivial cosmetic change that doesn't add meaningful value. Apologies for the noise.

PR #603, opened 2026-04-30
[BUG] Make GreedyEncoder.get_test_params a classmethod

### LINKED ISSUE #568: [BUG] GreedyEncoder.get_test_params should be a `classmethod` (skbase contract)

AbinaShah:
#### Reference Issues/PRs
Fixes #568.

- Make `GreedyEncoder.get_test_params` a `@classmethod` so it can be called at class-level (skbase estimator testing contract).
- Clarify `get_test_params` docstring return type and validate `parameter_set`.
- Add regression tests covering class-level invocation, parameter validity, and invalid `parameter_set` handling.

- API contract alignment with `skbase` expectations for `get_test_params`.
- Test coverage and whether the `parameter_set` validation behavior is appropriate.

- [x] Yes, added/updated tests in `pyaptamer/trafos/encode/tests/test_greedy_encoder.py`.

#### Any other comments?
- Ran focused tests locally: `pytest -q pyaptamer/trafos/encode/tests/test_greedy_encoder.py`.

#### PR checklist
- [x] Added/modified tests
- [x] Used pre-commit hooks when committing to ensure that code is compliant with hooks (or ran equivalent checks locally).

### DIFF
--- a/pyaptamer/trafos/encode/_greedy.py
+++ b/pyaptamer/trafos/encode/_greedy.py
@@ -132,7 +132,8 @@ class GreedyEncoder(BaseTransform):
 
         return result_df
 
-    def get_test_params(self):
+    @classmethod
+    def get_test_params(cls, parameter_set="default"):
         """Get test parameters for GreedyEncoder.
 
         Returns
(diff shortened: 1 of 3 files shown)


AbinaShah:
ran the focused unit tests for GreedyEncoder; full suite has unrelated network/HF-cache dependent failures in some environments.

PR #609, opened 2026-05-01
fix: PositionalEncoding Shape Mismatch on Odd Dimensions

### LINKED ISSUE #607: [BUG]Positional Encoding Shape Mismatch on Odd Dimensions
### 📋 Context
`PositionalEncoding` is a standard transformer component that injects sequence order information by adding sine and cosine waves of different frequencies to the embeddings.

### 🔍 The Bug Explained
The implementation uses a vector called `div_term` to calculate these frequencies. The logic is as follows:
- **Sine waves** are applied to **even** indices (`0, 2, 4...`)
- **Cosine waves** are applied to **odd** indices (`1, 3, 5...`)

The size of `div_term` is calculated as `ceil(d_model / 2)`. 
- If `d_model` is **128** (even): Both even and odd indices have **64** slots. `div_term` is size **64**. Everything matches.
- If `d_model` is **127** (odd): Even indices have **64** slots, but odd indices only have **63** slots. `div_term` is still size **64**.

When the code tries to assign the 64-value cosine result into the 63 odd-indexed slots:
```python
pe[0, :, 1::2] = torch.cos(position * div_term)
```
PyTorch throws a `RuntimeError` because it cannot fit 64 values into a space meant for 63.

### ⚠️ Why This Matters
While "power of 2" dimensions (like 128, 256) are common, many research experiments use custom embedding sizes. This bug makes the library fragile and prevents it from being a drop-in component for non-standard Transformer architectures.

### 🛠️ The Fix
The fix is to ensure that for the **odd** indices (the cosine part), we only use as many frequency terms as there are available slots. Slicing `div_term[:d_model // 2]` perfectly resolves the mismatch for any odd integer.

***

agam263:
***

#  Fix: PositionalEncoding Shape Mismatch on Odd Dimensions
fix #607 

## 📖 Summary
This PR resolves a persistent runtime crash in the `PositionalEncoding` layer that occurred whenever the model dimension (`d_model`) was an odd number. The fix ensures that the layer correctly handles arbitrary embedding dimensions, making the `AptaTrans` architecture more flexible and robust.

## 🔍 Technical Root Cause
The `PositionalEncoding` layer generates fixed sinusoids to inject positional information into embeddings. The implementation uses a frequency-divisor vector (`div_term`) that is shared between sine and cosine operations:

1.  **Index Split**: Sine waves are assigned to even indices (`0, 2, 4...`) and cosine waves to odd indices (`1, 3, 5...`).
2.  **The Mismatch**: 
    - The size of `div_term` is calculated based on `torch.arange(0, d_model, 2)`, which has a length of `ceil(d_model / 2)`.
    - If `d_model` is **128** (even): 
        - Even indices: 64 slots
        - Odd indices: 64 slots
        - `div_term`: 64 values (No error)
    - If `d_model` is **127** (odd):
        - Even indices: 64 slots
        - **Odd indices: 63 slots**
        - `div_term`: 64 values
3.  **The Crash**: The assignment `pe[0, :, 1::2] = torch.cos(position * div_term)` fails for odd `d_model` because PyTorch cannot fit 64 values into a tensor with 63 slots.

### 1. Corrected Frequency Mapping
Adjusted the cosine assignment to correctly slice the `div_term` to match the number of available odd-indexed slots.
- **Logic**: Used `div_term[: d_model // 2]` to ensure the number of frequencies exactly matches the number of odd positions, regardless of whether `d_model` is even or odd.

### 2. Dimensionality Regression Tests
Introduced `pyaptamer/aptatrans/tests/test_pe_robustness.py` to safeguard against future regressions:
- **Odd/Even Verification**: Tests both standard (even) and non-standard (odd) dimensions.
- **Extreme Edge Cases**: Verified stability for minimal dimensions (e.g., `d_model=1`).

## ⚠️ Impact & Risks
- **Architectural Flexibility**: Developers can now experiment with any embedding size without the model crashing during the initialization of the Transformer blocks.
- **Stability**: This is a safe change that does not modify the output for existing models where `d_model` is even.

## ✅ Verification Results
- **Unit Tests**: `pytest pyaptamer/aptatrans/tests/test_pe_robustness.py` -> **Passed**.
- **Inference Stability**: Verified that the change preserves identical output values for even-dimension embeddings.

### DIFF
--- a/pyaptamer/aptatrans/layers/_encoder.py
+++ b/pyaptamer/aptatrans/layers/_encoder.py
@@ -61,7 +61,9 @@ class PositionalEncoding(nn.Module):
         )
         pe = torch.zeros(1, max_len, d_model)  # Changed shape to (1, max_len, d_model)
         pe[0, :, 0::2] = torch.sin(position * div_term)  # Changed indexing
-        pe[0, :, 1::2] = torch.cos(position * div_term)  # Changed indexing
+        pe[0, :, 1::2] = torch.cos(
+            position * div_term[: d_model // 2]
+        )  # Changed indexing
         self.register_buffer("pe", pe)
 
     def forward(self, x: Tensor) -> Tensor:
--- a/pyaptamer/aptatrans/tests/test_pe_robustness.py
+++ b/pyaptamer/aptatrans/tests/test_pe_robustness.py
@@ -1,0 +1,32 @@
+import pytest
+
+from pyaptamer.aptatrans.layers._encoder import PositionalEncoding
+
+
+def test_positional_encoding_odd_dimensions():
+    """
+    Test that PositionalEncoding handles odd-numbered model dimensions.
+    Regression test for Issue where odd d_model caused shape mismatch crash.
+    """
+    # Test standard even dimension
+    try:
+        pe_even = PositionalEncoding(d_model=128, max_len=10)
+        assert pe_even.pe.shape == (1, 10, 128)
+    except Exception as e:
+        pytest.fail(f"PositionalEncoding failed on even d_model: {e}")
+
+    # Test odd dimension (previously crashed)
+    try:
+        pe_odd = PositionalEncoding(d_model=127, max_len=10)
+        assert pe_odd.pe.shape == (1, 10, 127)
+    except Exception as e:
+        pytest.fail(f"PositionalEncoding failed on odd d_model: {e}")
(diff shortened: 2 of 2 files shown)


direkkakkar319-ops:
hi @agam263 your pr looks similar to #527

probably correct as you patched on the older file

PR #532, opened 2026-04-26
Clean up AptaTrans FCS extraction and add reproduction proof

archittmittal:
This PR removes the unrelated PSeAAC package churn from the previous attempt and keeps the scope on the AptaTrans protein-word frequency investigation.

Summary:
- restores the public PSeAAC package implementation at `pyaptamer/pseaac/*`
- removes the duplicate `pyaptamer/trafos/pseaac` copy and its tests
- adds an FCS regression test against the bundled `protein_word_freq.csv`
- hardens `filter_words()` for empty vocabularies so empty-input fits do not emit warnings
- fixes a Python 3.9-compatible test path in `pyaptamer/pseaac/tests/test_pseaac.py`

Proof / validation:
- `filter_words()` reproduces the bundled protein word frequency filtering from `protein_word_freq.csv`
- focused test run passed: `pyaptamer/trafos/encode/tests/test_fcs.py` and `pyaptamer/pseaac/tests/test_pseaac.py`

Fixes #172

PR #412, opened 2026-04-14
[ENH] Implement FCSWordTransformer for protein word frequency extraction

archittmittal:
This PR implements the FCSWordTransformer as requested in #172. 

Key changes:
- Created  in  which fits a vocabulary by counting protein k-mers (up to length 3) and filtering by mean frequency.
- Integrated the transformer into the  module.
- Added comprehensive tests in .

Closes #172

PR #615, opened 2026-05-01
AptaTrans: add FCSWordTransformer and reproduction test

archittmittal:
Adds FCSWordTransformer and a regression test reproducing the bundled protein word frequencies (datasets/data/protein_word_freq.csv). This PR contains only the FCS implementation and minimal utils fixes; unrelated PSeAAC and workflow changes were intentionally excluded.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,70 @@
+# This workflow will upload a Python Package to PyPI when a release is created
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#publishing-to-package-registries
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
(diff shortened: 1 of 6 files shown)


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,70 @@
+# This workflow will upload a Python Package to PyPI when a release is created
```
siddharth7113, line 1: Why is this file being added? 
I don't think this is in scope of this PR.

siddharth7113:
Also I see you have made closed and made multiple PRs on the same issue, please try to do all changes here, since it is difficult to find prior interactions and extends reviewing work, I am referencing all previous PR here so , it could be referenced while reviewing.

#412, #532

.github/workflows/python-publish.yml
siddharth7113: I think you were asked similar things in #412 as well, please make sure you are only editing code which concerns the scope of PR.

PR #642, opened 2026-05-04
[ENH] Expose AptaNet hyperparameters and add configuration flexibility

### LINKED ISSUE #169: [ENH] Changes to AptaNet for better customization

pranavchoudhary-tech:
## Description
This Pull Request addresses issue #169 by fully exposing `AptaNet` hyperparameters. Previously, users could not directly configure parameters like the number of estimators or optimizer choices, limiting the flexibility of the network.

This update introduces explicit hyperparameters for `n_estimators`, `max_depth`, `optimizer`, `device`, and `weight_decay`, propagating them effectively throughout the model architecture.

## Related Issues
- Resolves #169

## Key Changes
- Added hyperparameters: `n_estimators`, `max_depth`, `optimizer`, `device`, and `weight_decay` to `AptaNet` configurations.
- Implemented robust integration tests (`test_hyperparameters.py`) to verify parameter propagation and model compilation.
- Included edge-case testing to ensure proper validation for all configurations.
- Cleaned up related documentation and type hints.

## Checklist
- [x] Code follows the project's style guidelines.
- [x] Integration and unit tests added for new hyperparameters.
- [x] All existing tests pass successfully.
- [x] Appropriate parameter validation implemented.

### DIFF
--- a/pyaptamer/aptanet/_feature_classifier.py
+++ b/pyaptamer/aptanet/_feature_classifier.py
@@ -29,6 +29,53 @@ class AptaNetClassifier(ClassifierMixin, BaseEstimator):
 
     The estimator is non-deterministic and only supports binary classification.
 
+    Parameters
+    ----------
+    input_dim : int or None, default=None
+        Size of the input layer in the neural net. If `None`, it is inferred
+        from the feature matrix shape by the underlying module.
+    hidden_dim : int, default=128
+        Number of units in each hidden layer of the neural net.
+    n_hidden : int, default=7
+        Number of hidden layers in the neural net.
+    dropout : float, default=0.3
+        Dropout probability used in the neural net.
+    max_epochs : int, default=200
+        Maximum number of training epochs for the neural net.
+    lr : float, default=0.00014
+        Learning rate for the optimizer.
+    alpha : float, default=0.9
+        Discounting factor (rho) for the squared-gradient moving average
+        in RMSprop. Only used when ``optimizer`` is ``torch.optim.RMSprop``.
+    eps : float, default=1e-08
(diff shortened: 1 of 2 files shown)


pranavchoudhary-tech:
@satvshr review and approve workflows

PR #612, opened 2026-05-01
[CRASH] Fix: add ValueError for invalid encoder_type and None input_dim

DZDasherKTB:
#### Reference Issues/PRs
Fixes #610. Fixes #611.

#### What does this implement/fix? Explain your changes.

**Fix 1 : `_model_lightning.py` (`AptaTransEncoderLightning.configure_optimizers`)**
The `if/elif` chain for `encoder_type` had no `else` branch. Any value other than
`"apta"` or `"prot"` left `params` undefined, causing a silent
`NameError: name 'params' is not defined` at training time with no hint of the actual
cause. Added an `else` branch that raises a clear `ValueError` listing the valid options.

**Fix 2 : `_aptanet_nn.py` (`AptaNetMLP.__init__`)**
When `use_lazy=False` and `input_dim=None`, the guard
`first_lazy = use_lazy and (input_dim is None)` evaluated to `False`, causing
`nn.Linear(None, hidden_dim)` to be constructed and raising a confusing `TypeError`
deep inside PyTorch with no pointer to the real cause. Added an explicit guard that
raises `ValueError` immediately with an actionable message.

#### What should a reviewer concentrate their feedback on?
- The `else` branch in `configure_optimizers` in `_model_lightning.py` , confirm the
  error message lists all valid `encoder_type` values exhaustively.
- The guard placement in `_aptanet_nn.py` , confirm it sits before `first_lazy`
  assignment so no partial construction occurs.

#### Did you add any tests for the change?
Yes, one test per fix:
- `test_aptatrans_lightning.py` : `test_configure_optimizers_invalid_encoder_type_raises`:
  confirms `AptaTransEncoderLightning(model, encoder_type="invalid")` raises `ValueError`.
- `test_aptanet.py` : `test_aptanet_mlp_no_input_dim_raises`:
  confirms `AptaNetMLP(use_lazy=False, input_dim=None)` raises `ValueError`.

#### Any other comments?
Both are crash-level bugs with no workaround short of patching source. The fixes are
minimal and self-contained, one `else` branch and one `if` guard.

> LLM Assisted, by Claude (Anthropic), reviewed and submitted by DZDasherKTB

Issue #610, opened 2026-05-01
[CRASH] NameError in AptaTransEncoderLightning.configure_optimizers() for unsupported encoder_type


DZDasherKTB:
## Summary
`configure_optimizers()` crashes with an uninformative `NameError` when `encoder_type` is any value other than `"apta"` or `"prot"`.

## File & Line
`encoder_type_model_lightning.py`, line ~313

## Root Cause
The `if/elif` chain assigns `params` only for `"apta"` or `"prot"`. Any other value skips both branches, leaving `params` undefined. The subsequent `torch.optim.Adam(params, ...)` call then raises `NameError: name 'params' is not defined` — with no informative message about the invalid input.

```python
if self.encoder_type == "apta":
    params = list(...)
elif self.encoder_type == "prot":
    params = list(...)
# no else → NameError if encoder_type is anything else
optimizer = torch.optim.Adam(params, ...)  # ← crashes here
```

The class accepts `encoder_type: str` with no validation, so this failure happens silently until training time.

## Expected Behavior
A `ValueError` should be raised immediately with a clear message listing valid encoder types.

## Proposed Fix
Add an `else` branch (consistent with the guard already present in `AptaTrans.forward_encoder()`):

```python
else:
    raise ValueError(
        f"Unsupported encoder_type '{self.encoder_type}'. "
        "Expected one of: 'apta', 'prot'."
    )
```

## Severity
**CRASH** — Reproducible at runtime. No workaround without patching the source.

siddharth7113:
Do we really need this? , my assumption is docstrings tell you which args are present for it? A good error is useful but feels excessive, leaving the final call for @NennoMP

NennoMP:
Seems reasonable to me, raising an exception for an unknown value of a "categorical" parameter , I would go ahead with a PR.

DZDasherKTB:
@NennoMP , I had made a pr closing this issue. kindly have a look: https://github.com/gc-os-ai/pyaptamer/pull/612

NennoMP:
#612 has a broader scope than this PR, including changes related to AptaNet. Please make separate PRs.

PR #660, opened 2026-05-09
[BUG] fix: 7 critical bugs across training, inference, and benchmarking pipelines

DZDasherKTB:
## Summary

Went through the entire pyaptamer codebase, aptanet, aptatrans, layers,
benchmarking, and utils, reading the architecture and understanding the full
training and inference flow before touching anything. Found 7 bugs ranging
from a silent architectural flaw affecting every trained model to hard crashes
at training time. All fixed in this single PR.

---

## Fix 1: AlphaDropout paired with wrong activation in AptaNetMLP

**File:** `pyaptamer/aptanet/_aptanet_nn.py`

**Problem:**
```python
# Before
return nn.Sequential(linear, nn.ReLU(), nn.AlphaDropout(dropout))
```
AlphaDropout is mathematically formulated to preserve the mean and variance
of SELU-activated networks (self-normalising property, Klambauer et al.
NeurIPS 2017). Pairing it with ReLU breaks these guarantees entirely. The
network still trains and produces numbers, the bug is completely silent,
but gradient flow is suboptimal and no self-normalising property holds.
Every model trained with this codebase was affected on every hidden layer.

**Fix:**
```python
# After
return nn.Sequential(linear, nn.SELU(), nn.AlphaDropout(dropout))
```

---

## Fix 2: configure_optimizers() crashes with UnboundLocalError

**File:** `pyaptamer/aptatrans/_model_lightning.py`

**Problem:**
```python
def configure_optimizers(self):
    if self.encoder_type == "apta":
        params = list(self.model.encoder_apta.parameters()) + ...
    elif self.encoder_type == "prot":
        params = list(self.model.encoder_prot.parameters()) + ...
    # No else branch: params is undefined for any other value
    optimizer = torch.optim.Adam(params, ...)  # UnboundLocalError here
```
Any encoder_type value other than "apta" or "prot" produces an
`UnboundLocalError: local variable 'params' referenced before assignment`
at training time, far from the actual mistake, with no helpful message.
Also fixed the docstring typo "wrapper fro Lightning".

**Fix:** Validation added to `__init__` so the error fires immediately
with a clear ValueError, and an else branch added in `configure_optimizers`
as a safety net.

---

## Fix 3 and 4: assert statements stripped in optimised builds

**Files:** `pyaptamer/aptatrans/layers/_encoder.py`,
`pyaptamer/aptatrans/layers/_interaction_map.py`

**Problem:**
```python
assert x.shape[1] <= self.max_len, (...)
assert x_apta.shape[-1] == x_prot.shape[-1], (...)
```
Python's `-O` flag strips all assert statements. In optimised builds both
validations silently disappear, producing cryptic index errors deep in
tensor operations with no pointer to the actual cause.

**Fix:**
```python
if x.shape[1] > self.max_len:
    raise ValueError(...)

if x_apta.shape[-1] != x_prot.shape[-1]:
    raise ValueError(...)
```
Also fixed `if self.dropout` in PositionalEncoding to `if self.dropout.p > 0`
since nn.Dropout is an object and is always truthy regardless of p.

---

## Fix 5: recommend() loops infinitely on low-diversity searches

**File:** `pyaptamer/aptatrans/_pipeline.py`

**Problem:**
```python
while len(candidates) < n_candidates:
    result = mcts.run(verbose=verbose)
    ...
    if candidate not in candidates:
        candidates[candidate] = ...
```
If MCTS converges and keeps generating the same candidate, which happens
on small sequence spaces or low-diversity searches, this loops forever
with no way to break out and no indication of what is happening.

**Fix:**
```python
max_attempts = self.max_recommendation_attempts or n_candidates * self.n_iterations * 10
attempts = 0

while len(candidates) < n_candidates:
    if attempts >= max_attempts:
        logger.warning(
            "recommend() exhausted %d attempts but found only %d/%d unique candidates.",
            max_attempts, len(candidates), n_candidates,
        )
        break
    attempts += 1
```
Added `max_recommendation_attempts` parameter to `__init__` for explicit
control. Returns partial results with a warning instead of hanging.

---

## Fix 6: Pretrained weights re-downloaded on every run

**File:** `pyaptamer/aptatrans/_model.py`

**Problem:**
```python
path = os.path.relpath(
    os.path.join(os.path.dirname(__file__), ".", "weights", "pretrained.pt")
)
```
`os.path.relpath` makes the path relative to the process's current working
directory, not the file's location. Running from any directory other than
the package root produces a wrong path, `os.path.exists` returns False,
and the weights download from HuggingFace every single run.

**Fix:**
```python
path = os.path.abspath(
    os.path.join(os.path.dirname(__file__), "weights", "pretrained.pt")
)
```

---

## Fix 7: Duplicate estimator names silently corrupt Benchmarking results

**File:** `pyaptamer/benchmarking/_base.py`

**Problem:**
```python
est_name = estimator.__class__.__name__
results[est_name] = est_scores
```
Passing two instances of the same class with different hyperparameters
causes the second to silently overwrite the first in the results dict.
The returned DataFrame has one row instead of two with no error or warning.

**Fix:** Auto-generate names with index suffix (`ClassName_0`, `ClassName_1`).
Added optional `estimator_names` parameter for explicit control. Added
uniqueness validation that raises immediately on duplicates.

---

**Author:** Dashpreet Singh | dashpreetsinghhanda@gmail.com
IIT Jammu, B.Tech CSE 2024-2028

### DIFF
--- a/pyaptamer/aptanet/_aptanet_nn.py
+++ b/pyaptamer/aptanet/_aptanet_nn.py
@@ -7,38 +7,45 @@ import torch.nn as nn
 def aptanet_layer(input_dim, output_dim, dropout, lazy=False):
     """
     Create a single AptaNet layer composed of a linear transformation,
-    ReLU activation, and AlphaDropout.
-
+    SELU activation, and AlphaDropout.
+ 
     Parameters
     ----------
     input_dim : int
         Size of each input sample. Ignored if `lazy=True`.
-
+ 
     output_dim : int
         Size of each output sample (i.e., number of neurons in the layer).
-
+ 
     dropout : float
         Dropout probability for AlphaDropout. Must be between 0 and 1.
-
+ 
     lazy : bool, optional
         If True, use `nn.LazyLinear` instead of `nn.Linear`, allowing the input
         size to be inferred at runtime. Default is False.
-
+ 
     Returns
     -------
     nn.Sequential
         A sequential container with:
         - Linear or LazyLinear layer
-        - ReLU activation
+        - SELU activation
         - AlphaDropout layer
+ 
+    Notes
+    -----
+    SELU and AlphaDropout are used together by design. AlphaDropout is
+    specifically formulated to preserve the mean and variance of SELU-activated
+    networks (self-normalising property). Using AlphaDropout with ReLU (as in
+    the original code) breaks these guarantees. See:
+    Klambauer et al., "Self-Normalizing Neural Networks", NeurIPS 2017.
     """
     linear = nn.LazyLinear(output_dim) if lazy else nn.Linear(input_dim, output_dim)
     return nn.Sequential(
         linear,
-        nn.ReLU(),
+        nn.SELU(),           # correct activation for AlphaDropout
         nn.AlphaDropout(dropout),
     )
-
 
 class AptaNetMLP(nn.Module):
     """
--- a/pyaptamer/aptatrans/_model.py
+++ b/pyaptamer/aptatrans/_model.py
@@ -266,8 +266,10 @@ class AptaTrans(nn.Module):
 
         If the weights are not found locally, they will be downloaded from hugging face.
         """
-        path = os.path.relpath(
-            os.path.join(os.path.dirname(__file__), ".", "weights", "pretrained.pt")
+        path = os.path.abspath(
+            os.path.join(
+                os.path.dirname(__file__),"weights","pretrained.pt"
+            )
         )
 
         if os.path.exists(path):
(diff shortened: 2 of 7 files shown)


siddharth7113:
Hi, 

None of these are critical bugs.

1. I think this has already been addressed in a PR #362 
2.  You have already opened issue #610 for it, and PR #612 please don't duplicate issues and PR
3. I think this is a compiler specific thing, not really an issue (commonly caught by LLM as a crititical bug)


You have mixed a lot of issues together, none of which I would say is crtical and some of them even are no longer relevant, If you are using an LLM please review the output of it , to be sure they are valid bugs.

PR #319, opened 2026-04-07
[ENH] `deepDNAshape` algorithm for `DeepAptamer`

pandey-ps:
#### Reference Issues/PRs
Fixes #110 , refer #98

#### What does this implement/fix? Explain your changes.
Introduces the `deepDNAshape` [algorithm](https://github.com/JinsenLi/deepDNAshape) pytorch support for `DeepAptamer` implementation, the whole algorithm is not implemented, only the model architecture for calls, the converted weights and a predictor file just to return to the `DeepAptamer` call [here](https://github.com/gc-os-ai/pyaptamer/blob/cfd1b410b8974b0594f00167531e8aa9c901605b/pyaptamer/utils/_deepaptamer_utils.py#L84).

#### Did you add any tests for the change?

to be added

#### PR checklist

- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [x] Added/modified tests
- [x] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.
  To run hooks independent of commit, execute `pre-commit run --all-files`

PR #729, opened 2026-07-12
[ENH] Add Torch deepDNAshape predictor for DeepAptamer

### LINKED ISSUE #110: [ENH] `deepDNAshape` reworked
For DeepAptamer, one of the required components is the feature extractor (deep)DNAshape.

As per discussion in https://github.com/gc-os-ai/pyaptamer/issues/82#issuecomment-3193622643 and following, a package exists
https://github.com/JinsenLi/deepDNAshape

however it is not directly usable for us:

* it is not a pypi installable package
* it is `tensorflow`, not `torch` based
* it requires a comparably ancient `numpy<1.24`

In order to enable the reimplementation of DeepAptamer, we have hence discussed that the only feasible way forward (with an algorithm that includes deepDNAshape) is forking the deepDNAshape package into our library.

In the process of this, translating to `torch` to remove the version conflicts.

Crediting and license is straightforward:

* license is BSD-3, permissive. We need to credit code that is copied or inspired by
* in docstrings as well, we whould mention the original author, @jinsenli to give clear credit, and cite the original publication

Alleny244:
## Reference Issues/PRs

Fixes #110. Builds on #319. See also #98.

---

## What does this implement/fix? Explain your changes.

This PR ports **deepDNAshape** from TensorFlow to PyTorch, enabling DeepAptamer to use DNA shape features without relying on the original TensorFlow implementation or its outdated NumPy dependencies.

### Highlights

- Added `pyaptamer.deepdnashape`
  - `DNAModel`
  - `deepDNAshape.predict`
- Loads pretrained model weights from Hugging Face (weights are **not** bundled with the package)
- Includes scaling parameters (`_params.json`) as package data
- Added documentation explaining:
  - intrabase (`N`) vs interbase (`N-1`) outputs
  - model architecture
  - weight loading
  - attribution to the original deepDNAshape project
- Renamed the custom GRU implementation to `DualBiasGRUCell` with improved documentation
- Added `huggingface_hub` as a dependency

This work builds on the earlier implementation in #319, with additional packaging, documentation, cleanup, and test improvements.

---

## What should a reviewer concentrate their feedback on?

- Whether the `deepDNAshape.predict` API is intuitive for DeepAptamer integration
- Documentation clarity, especially around:
  - intrabase vs interbase outputs
  - layer selection
  - model weight loading
- Whether the frozen numerical reference tests are sufficient, or if TensorFlow parity testing should be pursued
- Packaging changes:
  - `huggingface_hub` dependency
  - inclusion of `_params.json` as package data

---

## Did you add any tests for the change?

Yes.

`pyaptamer/deepdnashape/tests/test_deepdnashape.py` covers:

- Invalid feature and layer validation
- Intrabase (`N`) vs interbase (`N-1`) output shapes
- Layer selection smoke tests
- Reverse-complement invariance
- Frozen numerical reference values for:
  - MGW
  - ProT
  - Roll
  - HelT

---

## Any other comments?

The original **deepDNAshape** project is licensed under BSD-3, and the documentation credits Jinsen Li and the upstream repository.

Running TensorFlow-versus-PyTorch parity tests in CI is currently impractical because the upstream implementation depends on legacy TensorFlow and older NumPy versions. Instead, this PR locks the PyTorch implementation's numerical behavior using frozen reference fixtures. I'm happy to adjust the testing strategy based on reviewer feedback.

---

## PR Checklist

- [x] PR title starts with one of `[ENH]`, `[BUG]`, `[DOC]`, or `[MNT]`
- [x] Added or modified tests
- [x] Ran `pre-commit` hooks (`pre-commit run --all-files`)


siddharth7113:
@Alleny244  could you remove `uv.lock` from the diff, I think it is inflating the loc

Alleny244:
> @Alleny244 could you remove `uv.lock` from the diff, I think it is inflating the loc

Sorry, didnt notice that, have removed the same

pyaptamer/deepdnashape/_predictor.py
```diff
@@ -0,0 +1,299 @@
...
+    if method == "minmax2":
+        return (predictions + 1) * (params["max"] - params["min"]) / 2 + params["min"]
+    if method == "sin":
+        return np.arcsin(predictions) / np.pi * 180.0
+    if method == "standard":
+        return predictions * params["std"] + params["mean"]
+    return predictions * params["percentile_range"] + params["median"]
+
+
+class deepDNAshape:  # noqa: N801
```
siddharth7113: Instead of this `predict` Can you make it more towards the standard estimator contract with `fit/predict` from `BaseTransform`?

pyaptamer/deepdnashape/_predictor.py
```diff
@@ -0,0 +1,299 @@
...
+import json
+import os
+
+import numpy as np
+import torch
+from huggingface_hub import hf_hub_download
+
+from ._model import DNAModel
+
+_HF_REPO_ID = "parkneurals/deepdnashape"
```
siddharth7113: We need to put it in gc-os-ai org https://huggingface.co/gcos instead of personal account.

pyaptamer/deepdnashape/tests/test_deepdnashape.py
```diff
@@ -0,0 +1,158 @@
...
+# test sequence
+TEST_SEQ = "AGCTTAGCGTACAGCTTAAAAGGGTTTCCCCTGCCCGCGTAC"
+
+# Short sequence used for numerical reference checks.
+REF_SEQ = "AGCTTAGCGT"
+
+# Frozen Torch-port outputs for REF_SEQ at layer=4.
+# These lock current numerical behavior in CI. Re-check against the
+# original TensorFlow deepDNAshape if model weights or rescaling change.
+_REF_PREDICTIONS = {
```
siddharth7113, line 18: Why is this needed? and why are they in `np.array` , Wouldn't ultimately model convert into `float32`?

pyaptamer/deepdnashape/tests/test_deepdnashape.py
```diff
@@ -0,0 +1,158 @@
...
+            37.09857941,
+            33.07782745,
+            34.76763916,
+        ],
+        dtype=np.float64,
+    ),
+}
+
+
+@pytest.fixture
```
siddharth7113: can you remove this as a fixture and make direct calls in the test, makes it more redable since it's just intialising the class.

pyaptamer/deepdnashape/tests/test_deepdnashape.py
```diff
@@ -0,0 +1,158 @@
...
+    assert np.issubdtype(preds.dtype, np.floating)
+    assert preds.shape == (len(TEST_SEQ) - 1,)
+
+
+@pytest.mark.parametrize("layer", [0, 4, 7])
+def test_layer_prediction(predictor, layer):
+    """Test if valid layer numbers produce predictions successfully."""
+    feature = "MGW"
+    preds = predictor.predict(feature, TEST_SEQ, layer=layer)
+    assert preds.shape == (len(TEST_SEQ),)
```
siddharth7113: can we make this into 1 test itself since asserts are pretty much the same?

pyaptamer/deepdnashape/_predictor.py
```diff
@@ -0,0 +1,299 @@
...
+        for bp2 in bases:
+            di[(bp, "N")] += di[(bp, bp2)]
+            di[("N", bp)] += di[(bp2, bp)]
+        di[(bp, "N")] /= bits
+        di[("N", bp)] /= bits
+
+    return mono, di
+
+
+def _build_graph(x):
```
siddharth7113: Is this all happening on `cpu`, Shouldn't it be taking argument `device`? Can you clarify here?

pyaptamer/deepdnashape/_model.py
```diff
@@ -0,0 +1,375 @@
...
+            ]
+        )
+
+        self.dropout = nn.Dropout(dropout_rate)
+        self.avg_layer = AvgFeatures(base_features, filter_size)
+
+    def _call_avg(self, x):
+        if self.training:
+            x = self.dropout(x)
+        return self.avg_layer(x)
```
siddharth7113, line 322: what is this function doing here?

pyaptamer/deepdnashape/_params.json
```diff
@@ -0,0 +1 @@
...n": 5.02, "percentile_range": 4.629999999999999, "method": "robust"}, "HelT-FL": {"min": 1.62, "max": 49.83, "mean": 3.800106784384326, "std": 0.6095502291461872, "median": 3.79, "percentile_range": 1.8700000000000006, "method": "robust"}, "Shear-FL": {"min": 0.17, "max": 3.07, "mean": 0.2537409468806036, "std": 0.08391295267245921, "median": 0.25, "percentile_range": 0.10999999999999999, "method": "robust"}, "Stretch-FL": {"min": 0.08, "max": 3.53, "mean": 0.1274697841224005, "std": 0.06712241710568438, "median": 0.12, "percentile_range": 0.13, "method": "robust"}, "Buckle-FL": {"min": 4.48, "max": 153.93, "mean": 8.240416851028154, "std": 4.820329155400036, "median": 7.72, "percentile_range": 7.74, "method": "robust"}, "Opening-FL": {"min": 1.58, "max": 137.75, "mean": 3.417052700216576, "std": 5.1468003686141435, "median": 3.47, "percentile_range": 8.28, "method": "robust"}, "MGW-FL": {"min": 0.0, "max": 5.71, "mean": 1.0643126206914797, "std": 0.304793577300888, "median": 1.03, "percentile_range": 2.08, "method": "robust"}, "Stagger-FL": {"min": 0.17, "max": 2.14, "mean": 0.3190100370275495, "std": 0.065081007372701, "median": 0.32, "percentile_range": 0.2, "method": "robust"}}
```
siddharth7113: what is this comment here? please make it rathere a readme comment and ship it in data/ from root folder instead of shipping it here.

pyproject.toml
```diff
@@ -15,6 +15,7 @@ authors = [
 dependencies = [
     "biopython>=1.87",
     "datasets>=5.0.0",
+    "huggingface_hub",
```
siddharth7113: do we really need this separately I think `datasets` would automatically pull it

pyaptamer/deepdnashape/_predictor.py
```diff
@@ -0,0 +1,299 @@
...
+import numpy as np
+import torch
+from huggingface_hub import hf_hub_download
+
+from ._model import DNAModel
+
+_HF_REPO_ID = "parkneurals/deepdnashape"
+_PARAMS_PATH = os.path.join(os.path.dirname(__file__), "_params.json")
+
+_REV_COMPLEMENT = {"A": "T", "T": "A", "C": "G", "G": "C", "N": "N"}
```
siddharth7113: Instead of defining everything as constants on top, is it possible you can combine them or define in a single datastructure?

pyaptamer/deepdnashape/_predictor.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Returns
+    -------
+    tuple of (dict, dict)
+        (mono, di) where mono maps single-character bases to
+        np.ndarray of shape (4,) and di maps 2-tuples of
+        bases to np.ndarray of shape (16,).
+    """
+    bases = ["T", "G", "C", "A"]
+    bits = len(bases)
```
siddharth7113: what does `bits` mean here?

pyaptamer/deepdnashape/_model.py
```diff
@@ -0,0 +1,375 @@
...
+    Parameters
+    ----------
+    data : torch.Tensor, shape (E, C)
+        Values to sum.
+    segment_ids : torch.Tensor, shape (E,)
+        Segment index for each row of data.
+    num_segments : int
+        Total number of output segments.
+
+    Returns
```
siddharth7113: Do we really need this as a separate function, since here the graph is not arbitarty, and always combine `pair_prev` and `pairs_next` , if my understanding is correct, we could instead directly deal with it, instead of defining function separately, 

so something like

```python
idx = torch.arange(x.shape[0], device=x.device)
prev_sumx = x[(idx - 1).clamp(min=0)]       # left neighbour; ends look at themselves
next_sumx = x[(idx + 1).clamp(max=x.shape[0] - 1)]
````

File: pyaptamer/deepdnashape/_model.py [hunk 13]
```diff
@@ -0,0 +1,375 @@
...
+        Number of output scalars per node. Each scalar is the
+        mean of a group of channels. Default is 1.
+    filter_size : int, optional
+        Number of input channels to partition. Default is 64.
+    """
+
+    def __init__(self, target_features=1, filter_size=64):
+        super().__init__()
+        self.target_features = target_features if target_features != 0 else 1
+        self.pad_amount = filter_size % self.target_features
```
siddharth7113: I think the calculation here might be wrong , for eg. `target_feature=5` then ` pad amount = 64%5=4` then `group_size=64+4 // 5` while it would come out to 13, I think you would want to pad with `1` more , right? instead of adding 4 more?

pyaptamer/deepdnashape/_model.py
```diff
@@ -0,0 +1,375 @@
...
+        `AvgFeatures` reduction. Default is 1.
+    constraints : bool, optional
+        If True, collect an `AvgFeatures` prediction after
+        every `MessagePassingConv` layer and stack them
+        along a new axis. Default is True.
+    selflayer : bool, optional
+        If True, collect an `AvgFeatures` prediction from
+        the initial convolution output before any message
+        passing. Default is True.
+    multiply : str or None, optional
```
siddharth7113: After going through the code I think instead of `multiply` let's put 2 bool variable instead `gate` and `dual_weights`.

for eg.

```python

def __init__(self, filters=64, gate=True, dual_weights=True, bn_layer=True, gru_layer=True):
    ...
    if gate:
        if dual_weights:
            self.w_next_all = ...
            self.w_prev_all = ...
        self.b_all = ...

````

File: pyaptamer/deepdnashape/_predictor.py [hunk 15]
```diff
@@ -0,0 +1,299 @@
...
+              shape ``(N - 1,)``, one value per base step.
+        """
+        if feature not in _ALL_FEATURES:
+            raise ValueError(
+                f"Unknown feature. Must be one of {sorted(_ALL_FEATURES)}."
+            )
+        if not 0 <= layer <= 7:
+            raise ValueError(f"layer must be between 0 and 7, got {layer}.")
+
+        if feature not in self._models:
```
siddharth7113: I checked the repo but there were only 4 model weights but there are lot more features, Could you check why is there such a big discrepancy,.

siddharth7113 (changes requested):

pyaptamer/deepdnashape/_predictor.py
Alleny244: Sounds good! I don't have access to upload under the gcos org, so could you upload the weights there? Once the repo is available, I'll update _HF_REPO_ID to point to it.

pyaptamer/deepdnashape/_predictor.py
Alleny244: Sure

pyaptamer/deepdnashape/_model.py
Alleny244: Sure

pyaptamer/deepdnashape/_model.py
Alleny244: yeah you are right here, I have updated the code

pyaptamer/deepdnashape/_model.py
Alleny244: Yeah,I replaced them with clamp-based left/right neighbors. Also removed _build_graph / _segment_sum and simplified forward to take only x. Verified equivalence against the old path and the frozen numeric tests still pass

pyaptamer/deepdnashape/_predictor.py
Alleny244: Renamed to n_bases , it’s the alphabet size (len(bases) = 4) used for one-hot width, not binary bits.

pyaptamer/deepdnashape/_predictor.py
Alleny244: Yeah, I have added them into a single dict

pyproject.toml
Alleny244: Agreed, have removed it. datasets already pulls huggingface_hub, so a separate pin isn’t needed.

pyaptamer/deepdnashape/_params.json
Alleny244: Sure

pyaptamer/deepdnashape/_model.py
Alleny244: Helper for the shared “dropout (train only) → AvgFeatures” step; forward calls it in a few places to avoid duplication. At inference it just runs the average. I ll a doc string there

pyaptamer/deepdnashape/tests/test_deepdnashape.py
Alleny244: Yeah, Have Combined the intrabase / interbase / layer shape checks into a single parametrized test_output_shape since the asserts were the same.

pyaptamer/deepdnashape/tests/test_deepdnashape.py
Alleny244: Done

pyaptamer/deepdnashape/_predictor.py
Alleny244: _build_graph has since been removed as part of the clamp-based neighbor indexing refactor. The current inference path runs on CPU by default (torch.tensor() creates CPU tensors and the model is not moved with .to(device)). Internally, neighbor tensors are created with device=x.device, so they'll automatically follow the input device if we add GPU support later.

pyaptamer/deepdnashape/tests/test_deepdnashape.py
Alleny244: These are just the expected outputs for a fixed test sequence. We keep them so that if a future code change accidentally changes the model's predictions, the test will catch it. They're stored as np.array since that's what predict() returne and it makes the comparisons straightforward. The model still runs in float32, using float64 here is only to preserve the reference values in the test and doesn't affect inference.

pyaptamer/deepdnashape/_predictor.py
Alleny244: The original deepDNAshape supports many features, but the HF repo currently only has weights for MGW, ProT, Roll, and HelT (the DeepAptamer set). The larger feature list came from mirroring the upstream API/params

pyaptamer/deepdnashape/_model.py
siddharth7113: but do we have a training mode?, for this if my understanding is correct then this is mostly cosmetical at this point, instead can we rid of this and directly perofrm average inline during inference?

pyaptamer/deepdnashape/_predictor.py
siddharth7113: Can you bring this up as agenda point in today's meeting, I think we can ask franz directly to do that, meanwhile can you get the model weights on your personal hf, we can directly transfer it, I think matteo did this with aptatrans dataset.

Alleny244: Yeah sure thing, I ll add it in today's agenda

PR #732, opened 2026-07-13
[ENH] Adds sequence encoder transformer for AptaDiff diffusion preprocessing

### LINKED ISSUE #718: [ENH] AptaDiff: Sequence tokenization and one-hot transforms
We need to encapsulate the sequence tokenization and one-hot encoding for AptaDiff into a sklearn-compatible transformer (`BaseTransform`).

This issue brings the data processing onto the `_transform(X)` and `inverse_transform(X)` structure. The module needs to handle character-to-integer tokenization, target sequence padding, and conversion to one-hot tensors and back to strings.

aditi-dsi:
#### Reference Issues/PRs
closes #718 
depends on #752 

This PR introduces the `SequenceKHotEncoder` to preprocess raw biological sequences (for e.g., required by algorithms following similar preprocessing approach like AptaDiff diffusion pipeline). It standardizes fixed-length biological sequencs converting them into PyTorch K-hot tensors. Additionally, it includes an `inverse_transform` method to accurately decode model-generated tensors back into standard biological strings.

Yes, in `pyaptamer/trafos/tests/test_k_hot.py`
#### PR checklist

- [x] Added/modified tests


pyaptamer/aptadiff/tests/test_transforms.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+import torch
+
+from pyaptamer.aptadiff import AptamerOneHotEncoder
+from pyaptamer.data import MoleculeLoader
+
+APTAMER = "AGCTTAGCGTACAGCTTAAAAGGGTTTCCCCTGCCCGCGTAC"
+
+
+def test_aptamer_one_hot_encoder_from_moleculeloader():
```
siddharth7113: I would recommend to add test with a small fastq file which we could also ship as part of package using `load_fastq` or similar function to what we do with pdb.

relevant file details https://github.com/gc-os-ai/pyaptamer/issues/733

siddharth7113 (changes requested):

pyaptamer/aptadiff/tests/test_transforms.py
aditi-dsi: looking into it.

pyaptamer/aptadiff/tests/test_transforms.py
```diff
@@ -0,0 +1,113 @@
...
+
+import pandas as pd
+import pytest
+import torch
+
+from pyaptamer.aptadiff import AptamerOneHotEncoder
+from pyaptamer.data import MoleculeLoader
+
+APTAMER = "AGCTTAGCGTACAGCTTAAAAGGGTTTCCCCTGCCCGCGTAC"
+FASTQ_PATH = Path(__file__).parents[2] / "datasets" / "data" / "sample.fastq"
```
siddharth7113: instead of defining `FASTQ_PATH` do you think it would be better if we defined `load_sample_fastq` like function similar to `load_1gnh` and other functions we have?

pyaptamer/aptadiff/tests/test_transforms.py
```diff
@@ -0,0 +1,113 @@
...
+    """Column names are configurable, not hardcoded to aptamer."""
+    X = MoleculeLoader(data={"custom_seq": [APTAMER] * 2})
+    encoder = AptamerOneHotEncoder(target_length=40, aptamer_col="custom_seq")
+    Xt = encoder.fit_transform(X)
+
+    assert len(Xt) == 2
+    assert Xt.shape == (2, 40, 5)
+
+
+def test_aptamer_one_hot_encoder_inverse_transform():
```
siddharth7113, line 46: I think it is testing multiple things, all of which are valid tests , can you make separate unit tests for them?

pyaptamer/aptadiff/_transforms.py
```diff
@@ -0,0 +1,133 @@
...
+        -------
+        torch.Tensor
+            A 3D float32 tensor of shape (n_samples, target_length, num_classes).
+        """
+        encoded_seqs = []
+        pad_idx = self.vocab_[self.pad_token]
+
+        for seq in X[self.aptamer_col]:
+            seq = seq.upper()
+            seq = seq[: self.target_length]
```
siddharth7113: I just wanted to confirm this is what they are doing in original algorithm? Like they are truncating the sequences bigger than a certain length, instead of taking longest length and padding them?

pyaptamer/aptadiff/_transforms.py
```diff
@@ -0,0 +1,133 @@
...
+            indices = [self.vocab_.get(char, pad_idx) for char in seq]
+
+            if len(indices) < self.target_length:
+                indices += [pad_idx] * (self.target_length - len(indices))
+
+            encoded_seqs.append(indices)
+
+        int_tensor = torch.tensor(encoded_seqs, dtype=torch.long)
+        num_classes = len(set(self.inverse_vocab_))
```
siddharth7113: why is `set` needed here?

pyaptamer/aptadiff/tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Tests for the AptaDiff model"""
```
siddharth7113, line 1: add `_author` section here

pyaptamer/aptadiff/tests/test_transforms.py
```diff
@@ -0,0 +1,113 @@
+"""Tests for the AptamerOneHotEncoder transform."""
```
siddharth7113, line 1: Can you also add a test, about what would happen if I set pad_token="A"?

pyaptamer/aptadiff/_transforms.py
```diff
@@ -0,0 +1,133 @@
...
+        "output_type": "tensor",
+    }
+
+    def __init__(self, target_length=40, pad_token="P", aptamer_col="aptamer"):
+        self.target_length = target_length
+        self.pad_token = pad_token
+        self.aptamer_col = aptamer_col
+
+        self.vocab_ = {"A": 0, "T": 1, "G": 2, "C": 3, "U": 1, self.pad_token: 4}
+        self.inverse_vocab_ = {v: k for k, v in self.vocab_.items() if k != "U"}
```
siddharth7113: Can you move this to locally inside `_transform` and `inverse_transform` and removing the suffix `_` since it indicates it is learned in `fit`

aditi-dsi: on it.

pyaptamer/aptadiff/_transforms.py
aditi-dsi: The original authors are assuming a fixed length sequence. I kept truncation/padding to account for the cases if there comes a variable length data. 

Now, that we have confirmed that data will strictly be fixed length given by user (as per our discussion), I'll remove the truncation and padding process.

pyaptamer/aptadiff/_transforms.py
aditi-dsi: since `pad_token` logic is removed, so instead of duplicating the mappings inside both methods, wouldn't it be better to move these dicts to static class-level private constants outside `__init__`? 

Lmk if this looks better,

ref:
```python
_tags = {
        "property:fit_is_empty": True,
        "output_type": "tensor",
    }

    _VOCAB = {"A": 0, "T": 1, "G": 2, "C": 3, "U": 1}
    _INVERSE_VOCAB = {0: "A", 1: "T", 2: "G", 3: "C"}

    def __init__(self, aptamer_col="aptamer"):
        self.aptamer_col = aptamer_col
        super().__init__()
```

pyaptamer/aptadiff/_transforms.py
aditi-dsi: it's redundant. brain slip on my end.

pyaptamer/aptadiff/tests/test_transforms.py
aditi-dsi: yes, sounds better. I'm adding it inside `pyaptamer.datasets._loaders._fastq_loader`

pyaptamer/aptadiff/tests/test_transforms.py
aditi-dsi: padding has been removed, so this is outdated now.

siddharth7113 (changes requested):
Hi Aditi,

As discussed today please make the changes accordingly.

siddharth7113 (changes requested):
Since `capability:multivariate` is False, the base transformer rejects any X
with more than one column before `_transform` runs. So `sequence_col` can
only ever match the single column.


Please drop `sequence_col` and read the column by position, `X.iloc[:, 0]`,
as PrimerTrimmer does. For `inverse_transform`, record the input column name
in `_fit` (with `property:fit_is_empty` False) and reuse it, or return a
fixed name. The `get_test_params` override and the missing-column and
custom-column tests can be fixed then.

See #770 and the univariate rule now in the `_fit` / `_transform`
docstrings on main.

aditi-dsi:
Need a decision on the idea of renaming this encoder back to something along the lines of "one-hot" (`SequenceOneHotEncoder`, `SeqOneHotEncoder`, etc.).

#### Why?
In ML, "K-hot" or "multi-hot" generally implies a multi-label setup, a single V sized vector where multiple 1s exist in the same dimension and positional order is lost (like bag-of-words).

Our encoder doesn't output this. It returns a structured (V, L) matrix where every single position is a standard one-hot vector. 

Here is a vignette showing the output structure of one-hot vs k-hot for our case:
(Using name `OurEncoder` to avoid confusion as the name is not fixed yet)

```python
import pandas as pd
from pyaptamer.trafos.encode import OurEncoder

# A single sequence of length 4 (L=4)
X = pd.DataFrame({"reads": ["AACG"]})

encoder = OurEncoder(vocab={"A": 0, "C": 1, "G": 2, "T": 3})
Xt = encoder.fit_transform(X)

print(f"Shape: {Xt.shape}") 
# Shape: (1, 4, 4) -> (batch_size, vocab_size, seq_length)

print(Xt[0])
# Actual Output for "AACG":
# There's exactly one active bit per column:
#           1  2  3  4
#         -----------
#      A |  1  1  0  0  |         -> matrix of one-hot vectors
#      C |  0  0  1  0  |
#      G |  0  0  0  1  |
#      T |  0  0  0  0  |

# What a true K-Hot output would have looked like
# Output for "AACG":
#           A  C  G  T
#           -----------
#         [  1  1  1  0  ]         -> k-hot bag-of-words vector
```

Instead of collapsing the sequence into a single **K-hot bag-of-words vector,** our encoder produces a **matrix of one-hot vectors** for each sequence. Because each position retains its own isolated one-hot vector, the term "one-hot" sounds more applicable here than "k-hot".

aditi-dsi:
#### Need some architectural decision before merge:

While preparing for #778 I dig deeper into upstream implementation and observed that the one-hot encoding is not always needed in AptaDiff diffusion stage. For e.g., it is required during the internal processing by some functions like `log_sample_categorical`, `_train_loss` etc, which operate in log-one-hot space. While other places like `denoise_fn` operate directly on indices. And for this internal back-and-forth conversions we already have `index_to_log_onehot` and `log_onehot_to_index` functions introduced in #762. So, in that sense keeping this encoder meant strictly for seq-to-indices makes more sense, because anyway those internal helpers are going to handle the internal representations.

Just putting out the two possible options here to handle the representation at the transform vs model boundary: 

Option 1: SequenceEncoder outputs integer indices, and the diffusion model converts them to log-one-hot internally using `index_to_log_onehot()`.

```
SequenceEncoder -> integer indices -> AptaDiffDiffusion -> index_to_log_onehot() -> log-one-hot -> diffusion
````

Pros: matches the original AptaDiff implementation, keeps the model API simple, uses less memory, and keeps log-one-hot as an internal diffusion representation.

Option 2: SequenceEncoder outputs log-one-hot directly, so the diffusion model can avoid that initial conversion.

```
SequenceEncoder -> log-one-hot -> AptaDiffDiffusion -> use directly
```

Consequence: this saves only that one conversion, while the diffusion process still needs index representations internally for the embedding/denoising steps for which it will have to use the helper functions to convert back to indices. It also exposes transform to internal diffusion representations and vice versa.

I’m currently leaning toward Option 1 because it seems cleaner and closer to the original implementation, but wanted to get opinions.

siddharth7113:
> #### Need some architectural decision before merge:
> While preparing for #778 I dig deeper into upstream implementation and observed that the one-hot encoding is not always needed in AptaDiff diffusion stage. For e.g., it is required during the internal processing by some functions like `log_sample_categorical`, `_train_loss` etc, which operate in log-one-hot space. While other places like `denoise_fn` operate directly on indices. And for this internal back-and-forth conversions we already have `index_to_log_onehot` and `log_onehot_to_index` functions introduced in #762. So, in that sense keeping this encoder meant strictly for seq-to-indices makes more sense, because anyway those internal helpers are going to handle the internal representations.
> 
> Just putting out the two possible options here to handle the representation at the transform vs model boundary:
> 
> Option 1: SequenceEncoder outputs integer indices, and the diffusion model converts them to log-one-hot internally using `index_to_log_onehot()`.
> 
> ```
> SequenceEncoder -> integer indices -> AptaDiffDiffusion -> index_to_log_onehot() -> log-one-hot -> diffusion
> ```
> 
> Pros: matches the original AptaDiff implementation, keeps the model API simple, uses less memory, and keeps log-one-hot as an internal diffusion representation.
> 
> Option 2: SequenceEncoder outputs log-one-hot directly, so the diffusion model can avoid that initial conversion.
> 
> ```
> SequenceEncoder -> log-one-hot -> AptaDiffDiffusion -> use directly
> ```
> 
> Consequence: this saves only that one conversion, while the diffusion process still needs index representations internally for the embedding/denoising steps for which it will have to use the helper functions to convert back to indices. It also exposes transform to internal diffusion representations and vice versa.
> 
> I’m currently leaning toward Option 1 because it seems cleaner and closer to the original implementation, but wanted to get opinions.

Let;s do 1, doesn't make sense to do 2, especially if the transformer is just a `SequenceEncoder` and changing to log space would be a too specific of an operation, which is niche to one particular type of algorithm.

siddharth7113:
> Need a decision on the idea of renaming this encoder back to something along the lines of "one-hot" (`SequenceOneHotEncoder`, `SeqOneHotEncoder`, etc.).
> 
> #### Why?
> In ML, "K-hot" or "multi-hot" generally implies a multi-label setup, a single V sized vector where multiple 1s exist in the same dimension and positional order is lost (like bag-of-words).
> 
> Our encoder doesn't output this. It returns a structured (V, L) matrix where every single position is a standard one-hot vector.
> 
> Here is a vignette showing the output structure of one-hot vs k-hot for our case: (Using name `OurEncoder` to avoid confusion as the name is not fixed yet)
> 
> ```python
> import pandas as pd
> from pyaptamer.trafos.encode import OurEncoder
> 
> # A single sequence of length 4 (L=4)
> X = pd.DataFrame({"reads": ["AACG"]})
> 
> encoder = OurEncoder(vocab={"A": 0, "C": 1, "G": 2, "T": 3})
> Xt = encoder.fit_transform(X)
> 
> print(f"Shape: {Xt.shape}") 
> # Shape: (1, 4, 4) -> (batch_size, vocab_size, seq_length)
> 
> print(Xt[0])
> # Actual Output for "AACG":
> # There's exactly one active bit per column:
> #           1  2  3  4
> #         -----------
> #      A |  1  1  0  0  |         -> matrix of one-hot vectors
> #      C |  0  0  1  0  |
> #      G |  0  0  0  1  |
> #      T |  0  0  0  0  |
> 
> # What a true K-Hot output would have looked like
> # Output for "AACG":
> #           A  C  G  T
> #           -----------
> #         [  1  1  1  0  ]         -> k-hot bag-of-words vector
> ```
> 
> Instead of collapsing the sequence into a single **K-hot bag-of-words vector,** our encoder produces a **matrix of one-hot vectors** for each sequence. Because each position retains its own isolated one-hot vector, the term "one-hot" sounds more applicable here than "k-hot".

cc @fkiraly on this one, I am unable to make a decision at the moment.

aditi-dsi:
> > Need a decision on the idea of renaming this encoder back to something along the lines of "one-hot" (`SequenceOneHotEncoder`, `SeqOneHotEncoder`, etc.).
> > 
> > #### Why?
> > In ML, "K-hot" or "multi-hot" generally implies a multi-label setup, a single V sized vector where multiple 1s exist in the same dimension and positional order is lost (like bag-of-words).
> > 
> > Our encoder doesn't output this. It returns a structured (V, L) matrix where every single position is a standard one-hot vector.
> > 
> > Here is a vignette showing the output structure of one-hot vs k-hot for our case: (Using name `OurEncoder` to avoid confusion as the name is not fixed yet)
> > 
> > ```python
> > import pandas as pd
> > from pyaptamer.trafos.encode import OurEncoder
> > 
> > # A single sequence of length 4 (L=4)
> > X = pd.DataFrame({"reads": ["AACG"]})
> > 
> > encoder = OurEncoder(vocab={"A": 0, "C": 1, "G": 2, "T": 3})
> > Xt = encoder.fit_transform(X)
> > 
> > print(f"Shape: {Xt.shape}") 
> > # Shape: (1, 4, 4) -> (batch_size, vocab_size, seq_length)
> > 
> > print(Xt[0])
> > # Actual Output for "AACG":
> > # There's exactly one active bit per column:
> > #           1  2  3  4
> > #         -----------
> > #      A |  1  1  0  0  |         -> matrix of one-hot vectors
> > #      C |  0  0  1  0  |
> > #      G |  0  0  0  1  |
> > #      T |  0  0  0  0  |
> > 
> > # What a true K-Hot output would have looked like
> > # Output for "AACG":
> > #           A  C  G  T
> > #           -----------
> > #         [  1  1  1  0  ]         -> k-hot bag-of-words vector
> > ```
> > 
> > Instead of collapsing the sequence into a single **K-hot bag-of-words vector,** our encoder produces a **matrix of one-hot vectors** for each sequence. Because each position retains its own isolated one-hot vector, the term "one-hot" sounds more applicable here than "k-hot".
> 
> cc @fkiraly on this one, I am unable to make a decision at the moment.

I think this anyway has become stale as we have decided above to remove one-hot completely from the transform stage.

siddharth7113:
> > #### Need some architectural decision before merge:
> > While preparing for #778 I dig deeper into upstream implementation and observed that the one-hot encoding is not always needed in AptaDiff diffusion stage. For e.g., it is required during the internal processing by some functions like `log_sample_categorical`, `_train_loss` etc, which operate in log-one-hot space. While other places like `denoise_fn` operate directly on indices. And for this internal back-and-forth conversions we already have `index_to_log_onehot` and `log_onehot_to_index` functions introduced in #762. So, in that sense keeping this encoder meant strictly for seq-to-indices makes more sense, because anyway those internal helpers are going to handle the internal representations.
> > Just putting out the two possible options here to handle the representation at the transform vs model boundary:
> > Option 1: SequenceEncoder outputs integer indices, and the diffusion model converts them to log-one-hot internally using `index_to_log_onehot()`.
> > ```
> > SequenceEncoder -> integer indices -> AptaDiffDiffusion -> index_to_log_onehot() -> log-one-hot -> diffusion
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Pros: matches the original AptaDiff implementation, keeps the model API simple, uses less memory, and keeps log-one-hot as an internal diffusion representation.
> > Option 2: SequenceEncoder outputs log-one-hot directly, so the diffusion model can avoid that initial conversion.
> > ```
> > SequenceEncoder -> log-one-hot -> AptaDiffDiffusion -> use directly
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Consequence: this saves only that one conversion, while the diffusion process still needs index representations internally for the embedding/denoising steps for which it will have to use the helper functions to convert back to indices. It also exposes transform to internal diffusion representations and vice versa.
> > I’m currently leaning toward Option 1 because it seems cleaner and closer to the original implementation, but wanted to get opinions.
> 
> Let;s do 1, doesn't make sense to do 2, especially if the transformer is just a `SequenceEncoder` and changing to log space would be a too specific of an operation, which is niche to one particular type of algorithm.

Wait sorry I just have one more question, the covnersion should happen before going to the model right which is happening in second case but in first , why is integer indicies being passed to model (I assume) and then logone hot is applied. @aditi-dsi

siddharth7113:
Also take a look at this https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html we might be able to deign end to end at layer 3 with this too https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html

aditi-dsi:
> > > #### Need some architectural decision before merge:
> > > While preparing for #778 I dig deeper into upstream implementation and observed that the one-hot encoding is not always needed in AptaDiff diffusion stage. For e.g., it is required during the internal processing by some functions like `log_sample_categorical`, `_train_loss` etc, which operate in log-one-hot space. While other places like `denoise_fn` operate directly on indices. And for this internal back-and-forth conversions we already have `index_to_log_onehot` and `log_onehot_to_index` functions introduced in #762. So, in that sense keeping this encoder meant strictly for seq-to-indices makes more sense, because anyway those internal helpers are going to handle the internal representations.
> > > Just putting out the two possible options here to handle the representation at the transform vs model boundary:
> > > Option 1: SequenceEncoder outputs integer indices, and the diffusion model converts them to log-one-hot internally using `index_to_log_onehot()`.
> > > ```
> > > SequenceEncoder -> integer indices -> AptaDiffDiffusion -> index_to_log_onehot() -> log-one-hot -> diffusion
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Pros: matches the original AptaDiff implementation, keeps the model API simple, uses less memory, and keeps log-one-hot as an internal diffusion representation.
> > > Option 2: SequenceEncoder outputs log-one-hot directly, so the diffusion model can avoid that initial conversion.
> > > ```
> > > SequenceEncoder -> log-one-hot -> AptaDiffDiffusion -> use directly
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Consequence: this saves only that one conversion, while the diffusion process still needs index representations internally for the embedding/denoising steps for which it will have to use the helper functions to convert back to indices. It also exposes transform to internal diffusion representations and vice versa.
> > > I’m currently leaning toward Option 1 because it seems cleaner and closer to the original implementation, but wanted to get opinions.
> > 
> > 
> > Let;s do 1, doesn't make sense to do 2, especially if the transformer is just a `SequenceEncoder` and changing to log space would be a too specific of an operation, which is niche to one particular type of algorithm.
> 
> Wait sorry I just have one more question, the covnersion should happen before going to the model right which is happening in second case but in first , why is integer indicies being passed to model (I assume) and then logone hot is applied. @aditi-dsi

Because internally, different AptaDiff diffusion functions operate on `x` in either of two representations: indices or log-one-hot. If we hand over indices directly, we leave it to the diffusion layer to use the appropriate representation at each stage and handle the conversions internally with the existing helpers.

This is just a proposed change I am currently thinking of, the current version of encoder in this PR outputs one-hot.

aditi-dsi:
As discussed, I have updated the PR with final changes. Renamed the encoder back to the original name `SequenceOneHotEncoder`. Updated the test to follow the new name, improved docstrings and added some other minor optimizations. The CI is currently failing due to the issue outlined in #791, which I believe is introduced in the recent merge of #783.

siddharth7113:
> As discussed, I have updated the PR with final changes. Renamed the encoder back to the original name `SequenceOneHotEncoder`. Updated the test to follow the new name, improved docstrings and added some other minor optimizations. The CI is currently failing due to the issue outlined in #791, which I believe is introduced in the recent merge of #783.

I have replied on the issue, I have seen the failure and checking what is the best possible way to deal with `dna` , `rna` etc sequences in `test_base` i.e. unknow char issue, but the ohter is something you would have to resolved as responded [here](https://github.com/gc-os-ai/pyaptamer/issues/791#issuecomment-5759341197)

PR #739, opened 2026-07-16
[ENH] AptaDiff: Noise and denoise schedules, loss functions, transformer layers

### LINKED ISSUE #738: [ENH] AptaDiff: Noise, denoise schedules, loss functions, and transformer layers
We should implement the core tensor operations for the AptaDiff model, keeping the logic separated from the main estimator.

The goal is to add stateless PyTorch functions to handle the mathematical logic for:
- **Schedules:** Generating the noise arrays.
- **Noise & Denoise Steps:** Calculating how to add noise to the sequences (forward process) and how the model predicts the denoised steps (reverse process).
- **Loss Functions:** Calculating the training loss (ELBO), combining the KL divergence and cross-entropy reconstruction loss.

For layers, we should add:
- **Transformer Layers:** Positional Embedding and Linear Attention Transformer embedding layers. We should vendorize the original transformers implementation being used as the backend of the aforementioned two layers from the original author's [repository](https://github.com/lucidrains).

aditi-dsi:
#### Reference Issues/PRs

closes #738 

This PR introduces the core mathematical foundations for the AptaDiff diffusion operations. It currently implements the noise schedules, helper functions, and the forward diffusion transition probabilities, reverse step, VLB loss function, and transformer layers.
The following files contain the vendored parts of the custom frameworks ([Linear Attention Transformer](https://github.com/lucidrains/linear-attention-transformer) and [Axial Positional Embeddings](https://github.com/lucidrains/axial-positional-embedding)) used inside AptaDiff's transformer backend:
- `pyaptamer/aptadiff/layers/_axial_positional_embedding.py`
- `pyaptamer/aptadiff/layers/_local_attention.py`
- `pyaptamer/aptadiff/layers/_linear_attention_transformer.py`

Yes, inside:
- `pyaptamer/aptadiff/tests/test_functional.py`
- `pyaptamer/aptadiff/layers/tests/test_aptadiff_layers.py`
#### Any other comments?

no
#### PR checklist

- [x] Added/modified tests


pyaptamer/aptadiff/layers/_axial_positional_embedding.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import annotations
+
+__author__ = ["lucidrains"]
```
siddharth7113: Instead of putting original author put it as a comment on the top and mention the repo where it is taken from, along weith license and put yourself as author.

pyaptamer/aptadiff/layers/_linear_attention_transformer.py
```diff
@@ -0,0 +1,153 @@
+__author__ = ["lucidrains"]
```
siddharth7113: same here.

pyaptamer/aptadiff/layers/_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+    local_attn_window_size : int, default=128
+        Window size used for axial positional indexing and local attention.
+    transformer_type : str, default="native"
+        The attention backend to use.
+        - "native": Uses PyTorch's optimized nn.TransformerEncoderLayer.
+          Recommended for standard aptamer lengths (<= 100nt) as it provides exact
+          attention math and maximum hardware acceleration.
+        - "linear": Uses the linear attention approximation from the original
+          AptaDiff paper. Set to this for strict reproducibility.
+    **kwargs
```
siddharth7113: there is no backward comptability here?  Is this AI hallucination?

pyaptamer/aptadiff/layers/_libs/_axial_positional_embedding.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import annotations
```
siddharth7113, line 11: I think these files need to be added with a small `README.md` and explicilty link with LICENSE from upstream code, you can also confirm this with @fkiraly  about this in community meet.

pyaptamer/aptadiff/layers/_libs/_local_attention.py
```diff
@@ -0,0 +1,80 @@
...
+        B, H, n, dim_head = q.shape
+        b_flat = B * H
+
+        q = q.reshape(b_flat, n, dim_head)
+        k = k.reshape(b_flat, n, dim_head)
+        v = v.reshape(b_flat, n, dim_head)
+
+        scale = dim_head**-0.5
+        window_size = self.window_size
+        windows = n // window_size
```
siddharth7113, line 61: Add a value error here otherwise  this code would fail:

```python

import torch

from pyaptamer.aptadiff.layers._transformer import AptaDiffTransformerEmbedding

# A legal config: 128 % 64 == 0, so __init__ accepts it.
model = AptaDiffTransformerEmbedding(
    enc_embed_size=32,
    input_dim=4,
    output_dim=4,
    dim=64,
    depth=1,
    n_blocks=1,
    max_seq_len=128,
    local_attn_window_size=64,
    num_timesteps=1000,
    heads=8,
    n_local_attn_heads=4,  # with the default 0 the bug is invisible
    transformer_type="linear",
).eval()

t = torch.randint(0, 1000, (2,)).float()
z = torch.randn(2, 32)

for seq_len in (32, 63, 64, 100, 127, 128, 129):
    x = torch.randint(0, 4, (2, seq_len))
    try:
        with torch.no_grad():
            model(x, t, z)
        print(f"seq_len={seq_len:4d}  ({seq_len} % 64 = {seq_len % 64:2d})  OK")
    except Exception as e:
        print(
            f"seq_len={seq_len:4d}  ({seq_len} % 64 = {seq_len % 64:2d})  "
            f"{type(e).__name__}: {' '.join(str(e).split())[:70]}"
        )
```

siddharth7113: Output is like this:

```traceback
seq_len=  32  (32 % 64 = 32)  RuntimeError: shape '[1, 0, 64]' is invalid for input of size 32
seq_len=  63  (63 % 64 = 63)  RuntimeError: shape '[1, 0, 64]' is invalid for input of size 63
seq_len=  64  (64 % 64 =  0)  OK
seq_len= 100  (100 % 64 = 36)  RuntimeError: shape '[1, 1, 64]' is invalid for input of size 100
seq_len= 127  (127 % 64 = 63)  RuntimeError: shape '[1, 1, 64]' is invalid for input of size 127
seq_len= 128  (128 % 64 =  0)  OK
seq_len= 129  (129 % 64 =  1)  AssertionError: Sequence length (129) must be less than ...
```

siddharth7113: Can you in general insure all the asserts are present here as errors that were present in the upstream code.

pyaptamer/aptadiff/_functional.py
```diff
@@ -0,0 +1,281 @@
...
+    -------
+    torch.Tensor
+        Normalized log posterior for x_{t-1}. For t == 0, returns log_x0.
+    """
+    t_prev = torch.clamp(t - 1, min=0)
+
+    log_prior = q_forward(log_x0, t_prev, log_alphabar)
+    log_likelihood = q_forward_one_step(log_xt, t, log_alpha)
+    log_posterior = log_prior + log_likelihood
+    normalized_log_posterior = log_posterior - torch.logsumexp(
```
siddharth7113: This is a subtle bug but if you look at original implementation here
https://github.com/wz-create/AptaDiff/blob/ff2719c7840c18d2b912fd9fa4953ca5bd0f5602/diffusion_utils/diffusion_multinomial.py#L186-L206

here they are using `torch.where` on the **prior**, and then *still* multiplying by the
likelihood before normalising:

```python
log_EV_qxtmin_x0 = self.q_pred(log_x_start, t_minus_1)                           # the prior
log_EV_qxtmin_x0 = torch.where(t_broadcast == 0, log_x_start, log_EV_qxtmin_x0)  # <- swap is HERE
unnormed_logprobs = log_EV_qxtmin_x0 + self.q_pred_one_timestep(log_x_t, t)      # x likelihood
return unnormed_logprobs - torch.logsumexp(unnormed_logprobs, dim=1, keepdim=True)
```

Here we do the same `torch.where`, but three lines later, on the finished result:

```python
log_posterior = log_prior + log_likelihood
normalized_log_posterior = log_posterior - torch.logsumexp(log_posterior, dim=1, keepdim=True)
t_broadcast = t.view(-1, 1, 1)
return torch.where(t_broadcast == 0, log_x0, normalized_log_posterior)   # see here 

So at `t=0` we compute the likelihood and not use it,, the effect is that **at `t=0`
`q_posterior` ignores its own `log_xt` argument**.

File: pyaptamer/aptadiff/tests/test_functional.py [hunk 7]
```diff
@@ -0,0 +1,96 @@
...
+
+def test_posterior_t0_returns_x0_exactly(base_inputs):
+    """Verify q_posterior returns log_x0 exactly when t=0."""
+    log_x0, log_alpha, log_alphabar = base_inputs
+    log_xt = torch.log(torch.nn.functional.normalize(torch.rand(B, K, L), p=1, dim=1))
+    t = torch.zeros((B,), dtype=torch.long)
+
+    log_q = q_posterior(log_x0, log_xt, t, log_alpha, log_alphabar)
+
+    assert torch.equal(log_q, log_x0)
```
siddharth7113: this is validating the wrong behaviour, this might be AI going wrong here, because this is not the intended behaviour? Could you please recheck this?

pyaptamer/aptadiff/_functional.py
```diff
@@ -0,0 +1,281 @@
+__author__ = ["aditi-dsi"]
```
siddharth7113, line 3: Maybe add a top level docstring about what this file does

pyaptamer/aptadiff/_functional.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+def log_one_minus_exp(a: torch.Tensor) -> torch.Tensor:
+    """Computes log(1 - exp(a)) with numerical stability to avoid underflow."""
+    return torch.log(-torch.expm1(a))
+
+
+def log_add_exp(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
+    """Computes log(exp(a) + exp(b)) with numerical stability."""
+    return torch.logaddexp(a, b)
```
siddharth7113, line 17: maybe instead of making them as functions use it inline and add comments helps in understanding of code more easily.

pyaptamer/aptadiff/tests/test_functional.py
```diff
@@ -0,0 +1,96 @@
...
+
+    log_q = q_posterior(log_x0, log_xt, t, log_alpha, log_alphabar)
+    loss = compute_vlb_loss(log_x0, log_xt, t, log_alpha, log_alphabar, log_pred)
+
+    if t_val == 0:
+        expected_nll = -torch.sum(torch.exp(log_x0) * log_pred, dim=[1, 2])
+        assert torch.allclose(loss, expected_nll, atol=1e-5)
+    else:
+        expected_kld = torch.sum(multinomial_kl(log_q, log_pred), dim=1)
+        assert torch.allclose(loss, expected_kld, atol=1e-5)
```
siddharth7113, line 129: this is tatulogical you are using functions which are being tested to calculate the expected values and if they are wrong then the expected value would be wrong, but test would still pass.

for eg.

```python
def add(a, b):
    return a - b            # BUG

def test_add():
    expected = add(2, 3)    # ask the buggy code what the answer should be
    assert add(2, 3) == expected   # always passes
```

pyaptamer/aptadiff/tests/test_functional.py
```diff
@@ -0,0 +1,96 @@
...
+
+    log_pred = torch.log(torch.nn.functional.normalize(torch.rand(B, K, L), p=1, dim=1))
+    log_xt = torch.log(torch.nn.functional.normalize(torch.rand(B, K, L), p=1, dim=1))
+    t = torch.full((B,), t_val, dtype=torch.long)
+
+    log_q = q_posterior(log_x0, log_xt, t, log_alpha, log_alphabar)
+    loss = compute_vlb_loss(log_x0, log_xt, t, log_alpha, log_alphabar, log_pred)
+
+    if t_val == 0:
+        expected_nll = -torch.sum(torch.exp(log_x0) * log_pred, dim=[1, 2])
```
siddharth7113, line 125: this should not be calculated in a test, Could you please show an example of this test case failing, without changing the test?

pyaptamer/aptadiff/tests/test_functional.py
```diff
@@ -0,0 +1,96 @@
...
+    t = torch.full((B,), t_val, dtype=torch.long)
+
+    log_q = q_posterior(log_x0, log_xt, t, log_alpha, log_alphabar)
+    prob_q = torch.exp(log_q)
+
+    expected_ones = torch.ones((B, L))
+    assert torch.allclose(torch.sum(prob_q, dim=1), expected_ones, atol=1e-5)
+
+
+def test_posterior_t0_returns_x0_exactly(base_inputs):
```
siddharth7113: I couldn't understand this test, could you please show an example where this test would fail without changing anything in test?

pyaptamer/aptadiff/_functional.py
```diff
@@ -0,0 +1,281 @@
...
+
+    Returns
+    -------
+    torch.Tensor
+        Normalized log posterior for x_{t-1}. For t == 0, returns log_x0.
+    """
+    t_prev = torch.clamp(t - 1, min=0)
+
+    log_prior = q_forward(log_x0, t_prev, log_alphabar)
+    log_likelihood = q_forward_one_step(log_xt, t, log_alpha)
```
siddharth7113, line 198: why is `num_classes` being not passed here?  shouldn't an arg be `num_classes` here which should be passed to these functions? Could you maybe make default 4 in q_posterior and then instead of making it default in log_prior and log_likelihood we pass it there.

pyaptamer/aptadiff/_functional.py
```diff
@@ -0,0 +1,281 @@
...
+    on the specific time step sampled for each sequence:
+    - If the sampled step is t > 0: Computes the KL divergence between the true
+      denoising posterior and the neural network's prediction.
+    - If the sampled step is t = 0: Computes a standard cross-entropy
+      reconstruction loss (Negative Log-Likelihood) against the clean sequence.
+
+    Parameters
+    ----------
+    log_x0 : torch.Tensor
+        The initial clean, log-encoded sequence data.
```
siddharth7113, line 254: can you confirm if `log_x0`  is a tensor encodining is basically a log of one hot vectors made from sequences? because if so I think that would break the current code at t=0 at L203: `return where(True, log_x0, nz)`

pyaptamer/aptadiff/tests/test_functional.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+B, K, L = 2, 4, 10  # standard shape: (batch, num_classes, seq_len)
+TIMESTEPS = 1000
+
+
+@pytest.fixture
+def base_inputs():
+    """Provides standard data and schedules for the tests."""
+    log_x0 = torch.log(torch.nn.functional.normalize(torch.rand(B, K, L), p=1, dim=1))
```
siddharth7113, line 24: Can you explain how does we came to this? like is this how upstream defined the input?

siddharth7113: if we are using torch.rand it would always be >0 , is that the case or should we improve this test?

pyaptamer/aptadiff/layers/_transformer.py
aditi-dsi: backward compatible here means compatible with the original upstream LinearAttention code we're adapting (which supports more params like `causal`, `ff_dropout`, etc.). Someone may mistakenly pass those params based on the original AptaDiff README, so `**kwargs` just stops it from crashing instead of supporting them yet (as discussed earlier, we might eventually support them in future).

but i think moving `**kwargs` directly to pipeline later might make more sense. we can stop the extra params right there before reaching to this transformer layer. let me know what you think.

pyaptamer/aptadiff/layers/_libs/_local_attention.py
aditi-dsi: yes, it got dropped during vendoring, I'll add it back as ValueError.

pyaptamer/aptadiff/_functional.py
aditi-dsi: confirmed the bug due to one-line misplacement, will fix it to match upstream.

pyaptamer/aptadiff/tests/test_functional.py
aditi-dsi: this tests the same t=0 behavior that I need to fix in `q_posterior` (t=0 returns exact x0 instead of x0 with likelihood). Will update this test to match the corrected output once that bug is fixed.

pyaptamer/aptadiff/_functional.py
aditi-dsi: these are each used in two places, so keeping them as named functions avoids duplicating the formula (and inline comments) and names the formula explicitly (`log_one_minus_exp`, `log_add_exp`) for readability.

pyaptamer/aptadiff/tests/test_functional.py
aditi-dsi: This test still checks the thing it's meant to (correct branching between NLL and KL), so it's not wrong for that. But I think the real problem here is that `multinomial_kl` has no independent test, so a bug there wouldn't be caught. Will add a test for it.

pyaptamer/aptadiff/tests/test_functional.py
aditi-dsi:
```python
import torch
from pyaptamer.aptadiff._functional import cosine_alpha_schedule, q_posterior, multinomial_kl

B, K, L = 2, 4, 10
torch.manual_seed(0)
log_x0 = torch.log(torch.nn.functional.normalize(torch.rand(B, K, L), p=1, dim=1))
log_alpha = torch.log(cosine_alpha_schedule(1000))
log_alphabar = torch.cumsum(log_alpha, dim=0)
log_xt = torch.log(torch.nn.functional.normalize(torch.rand(B, K, L), p=1, dim=1))
log_pred = torch.log(torch.nn.functional.normalize(torch.rand(B, K, L), p=1, dim=1))
t = torch.zeros((B,), dtype=torch.long)

expected_nll = -torch.sum(torch.exp(log_x0) * log_pred, dim=[1, 2])  # test's own formula

def compute_vlb_loss_BUGGY(log_x0, log_xt, t, log_alpha, log_alphabar, log_pred):
    log_q = q_posterior(log_x0, log_xt, t, log_alpha, log_alphabar)
    kld = torch.sum(multinomial_kl(log_q, log_pred), dim=1)
    decoder_nll = torch.sum(torch.exp(log_x0) * log_pred, dim=[1, 2])  # bug: dropped the -ve sign
    mask = (t == 0).float()
    return mask * decoder_nll + (1.0 - mask) * kld

loss_buggy = compute_vlb_loss_BUGGY(log_x0, log_xt, t, log_alpha, log_alphabar, log_pred)
print("expected_nll:", expected_nll)
print("loss_buggy:  ", loss_buggy)
print("allclose?    ", torch.allclose(loss_buggy, expected_nll, atol=1e-5))
```
output:
```
expected_nll: tensor([15.1411, 15.2616])
loss_buggy:   tensor([-15.1411, -15.2616])
allclose?     False
```

pyaptamer/aptadiff/tests/test_functional.py
aditi-dsi: see here [comment](https://github.com/gc-os-ai/pyaptamer/pull/739#discussion_r3827572917)

pyaptamer/aptadiff/_functional.py
aditi-dsi: since `q_posterior` don't use `num_classes` directly, and it only affects `q_forward` and `q_forward_one_step`'s internal default, so instead of an explicit param, could derive it from `log_x0.shape[1]` there and skip the passing down entirely. thoughts?

pyaptamer/aptadiff/_functional.py
aditi-dsi: the flow from raw sequences to `log_x0` looks like this:

```
raw sequence strings
  -> AptamerOneHotEncoder (done in PR#732) -> outputs one-hot tensor, plain floats, shape (B,K,L)
  -> _index_to_log_onehot (will be done in ongoing PR#762) -> will do log conversion and clamping producing log_x0
  -> log_x0       -> this is fed into q_posterior/compute_vlb_loss (this PR)
```

so adhering to upstream code, the `.clamp(min=1e-30)` will be handled in the `_index_to_log_onehot` in pr #762 before passing to these functions.

pyaptamer/aptadiff/tests/test_functional.py
aditi-dsi: > Can you explain how does we came to this? like is this how upstream defined the input?

yes, see [here](https://github.com/wz-create/AptaDiff/blob/master/diffusion_utils/diffusion_multinomial.py#L61-L72)



> if we are using torch.rand it would always be >0 , is that the case or should we improve this test?

not guaranteed, torch.rand can technically hit exact 0, though that's very rare. But will add a fixture for the near-one-hot case, with the non-target classes clamped like real one-hot data (actual clamping will be done in model PR).

siddharth7113 (approved):
I think this looks good now, Should we merge it?

aditi-dsi:
> I think this looks good now, Should we merge it?

yes

siddharth7113:
Thanks done! good work @aditi-dsi

PR #741, opened 2026-07-17
 [ENH] Port RaptGen core components to layers

### LINKED ISSUE #717: Porting RaptGen core components to layers
Refs #715 
Went over `models.py` and `data.py` . Here is what I think can go into `pyaptamer/raptgen/layers/`:
- From `models.py`:
-`EncoderCNN`, `Bottleneck`, `DecoderPHMM`, `VAE`, `CNN_PHMM_VAE`, `kld_loss`, `State`, `Transition`, `profile_hmm_loss` and `profile_hmm_loss_fn`.
-`DecoderPHMM_fast` / `CNN_PHMM_VAE_FAST` / `profile_hmm_loss_fn_fast` have the same math as the previous ones, just reparametrized for speed, maybe we can include both versions .
- From  `data.py` : 
-`ProfileHMMSampler` , the rest of the code from this file are data loaders or dead code that we dont need.

I will port from the same original code from the raptgen repo not Aptadiff version.Difference between the two is mostly some helper functions and the extra `mu_filter` kwarg on `profile_hmm_loss_fn` used
in evaluation. .Can revisit later if some needed adjustments came up.

NoorMajdoub:
Closes #717

### What this does
Ports the RaptGen components  #717 into `pyaptamer/raptgen/`, following the same `_model.py` , `layers/` and tests structure used in `aptatrans`.

From `raptgen/models.py`:
- `pyaptamer/raptgen/layers/_conv.py`  : `Bottleneck`
- `pyaptamer/raptgen/layers/_encoder.py`  : `EncoderCNN`
- `pyaptamer/raptgen/layers/_decoder.py` : `DecoderPHMM`, `DecoderPHMM_fast`
- `pyaptamer/raptgen/layers/_loss.py` : `profile_hmm_loss`, `profile_hmm_loss_fn`, `profile_hmm_loss_fn_fast`, `torch_multi_polytope_dp_log`, `kld_loss`, `State`, `Transition`
- `pyaptamer/raptgen/layers/_utils.py` : `View`, `nt_index`
- `pyaptamer/raptgen/_model.py` : `VAE`, `CNN_PHMM_VAE`, `CNN_PHMM_VAE_FAST`

From `raptgen/data.py`:
- `pyaptamer/raptgen/layers/_sampler.py` : `ProfileHMMSampler`

No architectural or logic changes from (`hmdlab/raptgen`)

### Tests
Added tests check structural/shape logic (layer types/dimensions...)

#### PR checklist

- [x ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [x ] Added/modified tests
- [ ] Used pre-commit hooks when committing to ensure that code is compliant with hooks. Install hooks with `pre-commit install`.


siddharth7113 (dismissed):
Hi @NoorMajdoub,

I haven't reviewed the complete PR but there are some immediate issues that standout please correct these :

1.  pytests records file starting with `test_*` , here that is `_test` hence pytest wouldn't recognise it, also make it `tests` folder.
2.  In `tests/_init_.py` you don't need to expose all tests and simple exsistence of `_init_.py` should be okay, same goes for test file you don't need to expose it using `_all_`
3. Docsstrings are missing almost everywhere, please add correct numpydocstrings to all public classes/methods

siddharth7113:
Hi @NoorMajdoub ,

make sure your `pre-commit` is installed, it would automatically run when you make commits.

fkiraly:
> pytests records file starting with test_* , here that is _test hence pytest wouldn't recognise it, also make it tests folder.

Minor thing, I think pytest also recognizes files *ending* in the string "test".

pyaptamer/raptgen/layers/_conv.py
```diff
@@ -0,0 +1,70 @@
+"""Convolutional layers(Bottleneck) for RaptGen's variational autoencoder"""
+
+__author__ = ["NoorMajdoub"]
+__all__ = ["Bottleneck"]
+
+from torch import nn
+from torch.nn import functional as F
+
+
+class Bottleneck(nn.Module):
```
siddharth7113: Can we rename this to something that is more intuitive and easier to understand than `Bottleneck` 😅

pyaptamer/raptgen/layers/_conv.py
```diff
@@ -0,0 +1,70 @@
...
+        self.conv3 = nn.Conv1d(
+            in_channels=init_dim * 2, out_channels=init_dim, kernel_size=1
+        )
+
+        self.bn1 = nn.BatchNorm1d(init_dim)
+        self.bn2 = nn.BatchNorm1d(init_dim * 2)
+        self.bn3 = nn.BatchNorm1d(init_dim * 2)
+
+    def forward(self, input):
+        x = self.conv1(F.leaky_relu(self.bn1(input)))
```
siddharth7113, line 78: Can you also mention this is pre-activation Resenet implementation instead of standard implementation in docstring and cite this as reference https://arxiv.org/pdf/1603.05027 in docstrings, and this is the intended choice.

pyaptamer/raptgen/layers/_conv.py
```diff
@@ -0,0 +1,70 @@
...
+    conv3 : nn.Conv1d
+        1x1 convolution projecting back down to init_dim channels.
+    bn1, bn2, bn3 : nn.BatchNorm1d
+        Batch normalization applied before each convolution
+
+    """
+
+    def __init__(self, init_dim=32, window_size=7):
+        super().__init__()
+        assert window_size % 2 == 1, f"window size should be odd, given {window_size}"
```
siddharth7113: Can you raise here instead value error?

```
if window_size % 2 != 1:
    raise ValueError(f"`window_size` must be odd, but got {window_size}.")
```

pyaptamer/raptgen/layers/_utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+__author__ = ["NoorMajdoub"]
+__all__ = ["View", "nt_index", "State", "Transition"]
+
+from enum import IntEnum
+
+from torch import nn
+
+
+def one_hot_index(seq):
```
siddharth7113: The name of the function and intended functionality is confusing , if this is supposed to convert to str to int then replace the current name , it currently isn't performing one hot encoding.

pyaptamer/raptgen/layers/_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    M2M = 0
+    M2I = 1
+    M2D = 2
+    I2M = 3
+    I2I = 4
+    D2M = 5
+    D2D = 6
+
+
+class View(nn.Module):
```
siddharth7113: Can you specify where this is being used, I think both decoders have their own local copy and also the shape is different here from `View` present there. Is this intended?
```python
self.shape = (shape,)      # here
self.shape = shape         
```

pyaptamer/raptgen/layers/_decoder.py
```diff
@@ -0,0 +1,162 @@
...
+            nn.BatchNorm1d(hidden_size),
+            nn.LeakyReLU(negative_slope=0.01, inplace=True),
+        )
+
+        self.tr_from_M = nn.Sequential(
+            nn.Linear(hidden_size, hidden_size),
+            nn.LeakyReLU(negative_slope=0.01, inplace=True),
+            nn.Linear(hidden_size, (motif_len + 1) * 3),
+            nn.LeakyReLU(negative_slope=0.01, inplace=True),
+            View((-1, motif_len + 1, 3)),
```
siddharth7113: Can you check if this would be equvialent as `nn.Unflatten` https://docs.pytorch.org/docs/2.13/generated/torch.nn.modules.flatten.Unflatten.html

siddharth7113: We could get rid of `View` classes completely.

pyaptamer/raptgen/layers/_utils.py
```diff
@@ -0,0 +1,64 @@
+"""Utility functions and classes for RaptGen layers"""
+
+__author__ = ["NoorMajdoub"]
+__all__ = ["View", "nt_index", "State", "Transition"]
```
siddharth7113: one_hot_index is missing here, if it is a prviate method you can make it start with _ otherwise it should be added here.

pyaptamer/raptgen/layers/_encoder.py
```diff
@@ -0,0 +1,64 @@
...
+            Stack of `num_layers` `Bottleneck` blocks.
+    """
+
+    def __init__(self, embedding_dim=32, window_size=7, num_layers=6):
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.window_size = window_size
+
+        self.embed = nn.Embedding(
+            num_embeddings=4,  # [A,T,G,C]
```
siddharth7113: Would it not take special tokens i;.e `PAD`,`SOS`,`EOS`?

siddharth7113: maybe make it `len(nt_index)`

pyaptamer/raptgen/layers/_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    """Nucleotide-to-integer index mapping."""
+
+    A = 0
+    T = 1
+    G = 2
+    C = 3
+    PAD = 4
+    SOS = 5
+    EOS = 6
+    U = 1
```
siddharth7113, line 24: Here assigning U and T both 1 , Wouldn't that cause issue?

pyaptamer/raptgen/layers/_conv.py
```diff
@@ -0,0 +1,70 @@
...
+
+class Bottleneck(nn.Module):
+    """
+    1D convolutional residual bottleneck block.
+
+    A residual block used in the RaptGen CNN encoder for feature extraction, projects
+    the input into higher dimension space via a 1x1 convolution layer to capture the
+    features with a 7x7 convolution layer (7x7 by default, configurable via
+    window_size) then project back down into the initial dimension.
+    Parameters
```
siddharth7113, line 27: nit: Can you add a line before this.

siddharth7113: same goes for `decoder.py` and `sampler.py`

pyaptamer/raptgen/layers/_encoder.py
```diff
@@ -0,0 +1,64 @@
...
+            embedding_dim=embedding_dim,
+        )
+
+        modules = [Bottleneck(embedding_dim, window_size) for _ in range(num_layers)]
+        self.resnet = nn.Sequential(*modules)
+
+    def forward(self, seqences):
+        # change X from (N, L) to (N, L, C)
```
siddharth7113: I am assuming a standard convention i.e. N-> batch_size, L =length, C=channels but would be good to spell out as a comment.

pyaptamer/raptgen/layers/_encoder.py
```diff
@@ -0,0 +1,64 @@
...
+
+    num_layers : int, optional, default=6
+        Number of stacked `Bottleneck` residual blocks.
+
+    Attributes
+    ----------
+        embed : nn.Embedding
+            Learned nucleotide embedding of 4 tokens (A, T, G, C).
+
+        resnet : nn.Sequential
```
siddharth7113: resnet might be inaccurate name to this.

pyaptamer/raptgen/layers/_sampler.py
```diff
@@ -0,0 +1,229 @@
...
+        if not sequence_only:
+            return states, seq
+        else:
+            return seq
+
+    def calc_seq_proba(self, seq: str):
+        """
+        Compute the log-probability that this profile HMM generates the given seq.
+        """
+        one_hot_seq = torch.tensor(one_hot_index(seq))
```
siddharth7113: Here when a sequence like this is passed `one_hot_index('tatATGTTcG')` it would raise KeyError , due to insertion and deletion branch adding `atgc` and `_`

pyaptamer/raptgen/layers/_sampler.py
```diff
@@ -0,0 +1,229 @@
...
+        model_len = self.e.shape[0]
+        random_len = len(seq)
```
siddharth7113, line 169: same issue here a deletion would add `_` here it would be counted as  part of the aptamer.

pyaptamer/raptgen/layers/_sampler.py
```diff
@@ -0,0 +1,229 @@
...
+
+            if state == State.M:
+                # logger.info("{:.2f}, {:.2f}, {:.2f}, {:.2f}".format(*self.e[idx-1]))
+                p = np.copy(self.e[idx - 1])
+                p[np.argmax(p)] += 100000
+                seq += np.random.choice(list("ATGC"), p=p / sum(p))
+            elif state == State.I:
+                seq += "N"
+            else:
+                seq += "_"
```
siddharth7113, line 157: if we are anways returning both states and  sequences , wouldn't it make sense to do something like this

```python
 if state == State.M:
     seq += np.random.choice(list("ATGC"), p=self.e[idx - 1])
elif state == State.I:
     seq += np.random.choice(list("ATGC"))     # capitals, not "atgc"
 else:
      pass                
```

pyaptamer/raptgen/layers/_encoder.py
```diff
@@ -0,0 +1,64 @@
...
+
+        self.embed = nn.Embedding(
+            num_embeddings=4,  # [A,T,G,C]
+            embedding_dim=embedding_dim,
+        )
+
+        modules = [Bottleneck(embedding_dim, window_size) for _ in range(num_layers)]
+        self.resnet = nn.Sequential(*modules)
+
+    def forward(self, seqences):
```
siddharth7113: typo: `sequences`

pyaptamer/raptgen/layers/_sampler.py
```diff
@@ -0,0 +1,229 @@
...
+            if idx == self.a.shape[0]:
+                break
+
+            if state == State.M:
+                # logger.info("{:.2f}, {:.2f}, {:.2f}, {:.2f}".format(*self.e[idx-1]))
+                p = np.copy(self.e[idx - 1])
+                p[np.argmax(p)] += 100000
+                seq += np.random.choice(list("ATGC"), p=p / sum(p))
+            elif state == State.I:
+                seq += "N"
```
siddharth7113, line 155: What does `N` represent here? and how is it process after this? I am honestly very confused about this working, Could you briefly describe how sample is supposed to work, it's i/o and intended functionality.

siddharth7113 (changes requested):
Hi @NoorMajdoub ,

I have left some changes, and there are still parts to review such as tests, but could you check coverage once and ensure test coverage is complete for the parts of code you are adding.

pyaptamer/raptgen/layers/_utils.py
siddharth7113: @NoorMajdoub  Could you specify this if the above is intended?

siddharth7113: As mentioned do we reallyt need this or `nn.flatten` should  be enough for this?

siddharth7113 (changes requested):
Hi, there are still unaddressed comments here, please fix those.

pyaptamer/raptgen/layers/_conv.py
NoorMajdoub: Changed the name to ReverseBottleneck, it is more accurate as what this class does is that it projects the features to a higher dimension to apply the convolution the expanded space then projects back to the initial dimension. I can also rename it to ExpansionBlock or ResidualInvertedBottleneck to highlight the residual connections.

pyaptamer/raptgen/layers/_utils.py
NoorMajdoub: The use of the function is to map the nucleotides to numbers (A =0 ,T=1 and so on) as specified in the nt_index class which doesn't follow a universal bioinformatics standard it is just this codebase's own internal convention. The name is confusing because it isn't really one_hot encoding like [1,0,0,0..] , I renamed it to seq_to_indices and indicated in the function description is maps a nucleotide sequence to a list of integers.

pyaptamer/raptgen/layers/_utils.py
NoorMajdoub: Yes , in the original code the View class was defined in a separate file and then redefined as internal class in DecoderPHMM and DecoderPHMM_fast, I removed it from there and just made it being imported from utils.py
About its usage , nn.Flatten only collapses input tensors dimensions and can't replicate the usage of View , for example how it is used in DecoderPHMM (View is only used in DecoderPHMM and its fast version) View((-1, motif_len + 1, 2)) , cant be replicated with nn.Flatten.

pyaptamer/raptgen/layers/_decoder.py
NoorMajdoub: Yes I think we can , the `View `class is only used in `DecoderPHMM `and its fast version 6 times, tested them all and both `View `and `nn.Unflatten` return the same output, will replcate `View `with `nn.Unflatten` here then.

pyaptamer/raptgen/layers/_encoder.py
NoorMajdoub: renamed it to `blocks`.

pyaptamer/raptgen/layers/_utils.py
NoorMajdoub: I think not because they can be used interchangeably. In the decoding/generation this isn't an issue because the sampler picks directly from the letters "ATGC" (the original sampler logic) . Here they are used in the encoding/mapping from the nucleotides to their embedding, so shouldn't cause an issue right?

siddharth7113:
@aditi-dsi  to check for the files Aptadiff needed.

pyaptamer/raptgen/layers/_sampler.py
```diff
@@ -0,0 +1,166 @@
...
+                            Transition.M2I.value,
+                            Transition.M2D.value,
+                        ]
+                    )
+                ]
+            elif state == State.I:
+                p = np.array(
+                    [
+                        self.a[idx][Transition.I2M.value],
+                        self.a[idx][Transition.I2I.value],
```
siddharth7113, line 133: Hi @NoorMajdoub ,

I might finally have the answer for why they made it 0, after doing some research and discussion with AI I could reach to this conclusion:

> Parts AI generated

`most_probable()` does not terminate on 11 of 20 randomly initialised decoders (2 second cap per call). The current code returns on 9.

The insert branch was changed from `[I2M, 0, 0]` to `[I2M, I2I, 0]` and the next state is `argmax` of that list. Insert does not increment `idx`. When `I2I > I2M` at a position, `argmax` returns insert, `idx` is unchanged, the list is recomputed with the same values, and the same branch runs again. The loop has no exit.

The path probability is the product of the transition probabilities along the path. Each consecutive insert multiplies by `I2I`, which is below 1:


`I2M`                 leave immediately
`I2I * I2M`           one repeat, smaller
`I2I * I2I * I2M`     two repeats, smaller again

So the most probable step out of insert is always match, for any values of I2M and I2I. `[I2M, 0, 0]` produced that result by making the other two entries zero. Please replace the comparison with the assignment:
```python
elif state == State.I:
    state = State.M
```
Please also replace `while True` with a bounded loop, for example `3 * model_len iterations`, and raise `RuntimeError` if the bound is reached. With the assignment above the loop terminates on all 20 decoders.

The docstring says "most likely sequence". A step-wise argmax is a greedy walk, not the most probable path through the model. Please change the docstring to say greedy. A Viterbi version as discussed previously wuld be a separate issue.

Please add a test that runs `most_probable()` on decoder output under a timeout.

pyaptamer/raptgen/layers/_utils.py
```diff
@@ -0,0 +1,44 @@
+"""Utility functions and classes for RaptGen layers"""
+
+__author__ = ["NoorMajdoub"]
+__all__ = ["nt_index", "State", "Transition", "seq_to_indices"]
+
+from enum import IntEnum
+
+
+def seq_to_indices(seq):
```
siddharth7113, line 9: Would this function be used later? I couldn't find right now where is it called? Or maybe it was missed could you please confirm once?

pyaptamer/raptgen/layers/_loss.py
```diff
@@ -0,0 +1,311 @@
...
+    ------
+    probabilities : torch.Tensor
+        log-probabilities of the given output tensor with
+        shape (`batch`,)
+
+    """
+    model_length = emission_proba.shape[1]
+    batch_size, string_length = output.shape
+
+    F = (
```
siddharth7113, line 208: F is not a good choice for variable especially since `torch.nn.functional` is also imported as F.

pyaptamer/raptgen/layers/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    "profile_hmm_loss_fn",
+    "profile_hmm_loss_fn_fast",
+    "torch_multi_polytope_dp_log",
+    "kld_loss",
+]
+
+from pyaptamer.raptgen.layers._conv import Inverted_Bottleneck
+from pyaptamer.raptgen.layers._decoder import DecoderPHMM, DecoderPHMM_fast
+from pyaptamer.raptgen.layers._encoder import EncoderCNN
+from pyaptamer.raptgen.layers._loss import (
```
siddharth7113, line 23: I think bunch of imported wrongly here? these are not defined in `_loss.py`

pyaptamer/raptgen/layers/tests/test_all_raptgen_layers.py
```diff
@@ -0,0 +1,189 @@
...
+
+    assert model.h2mu.out_features == embed_size
+    assert model.h2logvar.out_features == embed_size
+
+
+@pytest.mark.parametrize(
+    "motif_len, embed_size, hidden_size, kernel_size, batch_size, seq_len",
+    [(4, 8, 16, 5, 3, 20), (10, 16, 32, 7, 2, 40)],
+)
+def test_cnn_phmm_vae_forward(
```
siddharth7113, line 168: Could you separate tests for `_model.py` and put it above the folder for `_layers` so weh ave right location for each type of tests.

siddharth7113 (changes requested):
Hi @NoorMajdoub,

I have done a review and left some comments for changes. I think we now have a clearer answer on why transition probabilities are defined the way they are (see the comment on `_sampler.py`). Also there were some mistakes in `_loss.py`: `ce_loss`, `multi_categorical_loss_fn` and `end_padded_multi_categorical_loss_fn` are defined but never called. I would prefer to remove them here and implement them where there is need for it. This would help catch bugs as well as ensure there is test coverage, which a lot of the current code is still missing. Please use `pytest-cov` to measure the coverage and see if we can write tests to improve it here:
`pytest pyaptamer/raptgen --cov=pyaptamer.raptgen --cov-report=term-missing`

Also I think `calc_seq_prob` is removed from this code, I just wanted to confirm that was intended ,because I had those in my notes from last round of review.

pyaptamer/raptgen/layers/tests/test_all_raptgen_layers.py
NoorMajdoub: Moved test_cnn_phmm_vae_layers and test_cnn_phmm_vae_forward to tests_model.py test files in a seperate folder.

pyaptamer/raptgen/layers/__init__.py
NoorMajdoub: Yes State, Transition were actually in utils not loss , fixed that

pyaptamer/raptgen/layers/_loss.py
NoorMajdoub: Renamed it to alpha in `torch_multi_polytope_dp_log` and `profile_hmm_loss`

pyaptamer/raptgen/layers/_utils.py
NoorMajdoub: It is used in the pipeline code for the mapping (pr 761)

NoorMajdoub:
@siddharth7113 
This is the tests coverage I have for now 
image

Also about `calc_seq_prob` , it scores how probable a given complete sequence is under the model, which is what the still-missing decoding step (turning the _/N skeleton from most_probable() into a final real sequence) needs to actually pick a winner among candidates. I think I removed it temporarily while trying to see if I could combine that scoring logic directly into the decoding/generation step instead of keeping it separate, but it might not be the best approach.

PR #746, opened 2026-07-23
[BUG] Track fitted state, pin sklearn < 1.8 along with adding test for base class.

### LINKED ISSUE #735: [BUG] BaseTransform does not track fitted state, and is invisible to sklearn's tag system
Two defects in `pyaptamer/trafos/base/_base.py`, both surfaced while writing the first stateful transformer (`PairsToTokens`, #734). Both reproduce on existing transformers, so neither is new.

## 1. `fit()` never marks the object as fitted

`BaseTransform` inherits `is_fitted` / `check_is_fitted()` from skbase, but `fit()` never sets `_is_fitted` and `transform()` never checks it.

```python
from pyaptamer.aptanet import PairsToFeatures
from pyaptamer.data import MoleculeLoader

X = MoleculeLoader(data={"aptamer": ["AGCUUAGCGUAC"], "protein": ["ACDEFGHIKLMN"]})
t = PairsToFeatures().fit(X)          # fit succeeds

t.is_fitted                            # -> False
t.check_is_fitted()                    # -> NotFittedError, after a successful fit
```

Consequence: skbase's fitted-state API is unusable on any transformer. A transformer with fitted state that calls `transform()` before `fit()` raises a bare `AttributeError` on the missing attribute rather than `NotFittedError`.

Note for whoever picks this up: `fit()` has an early return for the `property:fit_is_empty` tag, so the two branches need thinking about together.

## 2. Transformers have no `__sklearn_tags__`

`BaseTransform` extends *skbase*'s `BaseEstimator`, not sklearn's, so sklearn cannot read tags off a transformer.

```python
from sklearn.utils._tags import get_tags
from pyaptamer.aptanet import PairsToFeatures

get_tags(PairsToFeatures())
# DeprecationWarning: PairsToFeatures object has no attribute __sklearn_tags__.
# ... The default are now used instead since retrieving tags failed.
# This warning will be replaced by an error in 1.8.
```

Currently a warning; sklearn says it becomes an **error in 1.8**. 
Reproduced on sklearn 1.7.2.

siddharth7113:
Closes #735 

- Adds `self._is_fitted=True` in `trafos/base.py` which ultimately inherits from `skbase` object.
- pinning `scikit-learn<1.8` for estimator and core-dependiecs like sktime to not break the pipeline.

### DIFF
--- a/pyaptamer/trafos/base/_base.py
+++ b/pyaptamer/trafos/base/_base.py
@@ -1,7 +1,10 @@
 """Base transformation class."""
 
+__author__ = ["fkiraly", "siddharth7113"]
+
 import pandas as pd
 from skbase.base import BaseEstimator
+from sklearn.utils import InputTags, Tags, TargetTags, TransformerTags
 
 from pyaptamer.data import MoleculeLoader
 
@@ -38,11 +41,12 @@ class BaseTransform(BaseEstimator):
         self : object
             Returns self.
         """
-        if self.get_tag("property:fit_is_empty", False):
-            return self
-
         X_inner, y_inner = self._check_X_y(X, y)
-        self._fit(X=X_inner, y=y_inner)
+
+        if not self.get_tag("property:fit_is_empty", False):
+            self._fit(X=X_inner, y=y_inner)
+
+        self._is_fitted = True
         return self
 
     def _fit(self, X, y=None):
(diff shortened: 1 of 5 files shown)


fkiraly (approved):
In-principle this looks good, including tests!

However I think the tests, if this is your intended framework, could lead to low coverage:

* currently, the test framework can only test a single instance of given class, leading to low parameter or missing branch coverage
* the scenario handling is a bit manual

Two ways to address this:

* using pytest fixtures that generate more instances via `get_test_params`
* using the test class from `scikit-base` which already provides a scaffolding - might be too heavy, a simple use case of this is in `skpro` (the `sktime` use case is much more complicated)

PR #752, opened 2026-08-05
[ENH] Add transform pipeline for primer removal from SELEX data

### LINKED ISSUE #744: removal of forward and reverse primer from fastq data as a transformation
potential methods:
- pattern matching based subsequence extraction

aditi-dsi:
#### Reference Issues/PRs

closes #744

This PR aims to add a reusable transformation pipeline for SELEX-derived nucleotide sequences to remove forward and reverse primers from the sequences.
The PR also adds a `fastq_loader` for loading `sample.fastq` file for testing purpose.

yes

#### Any other comments?

should be merged before #732

#### PR checklist

- [x] Added/modified tests


pyaptamer/datasets/_loaders/_fastq_loader.py
```diff
@@ -0,0 +1,19 @@
...
+    """Load the sample FASTQ file as a MoleculeLoader.
+
+    Returns
+    -------
+    loader : MoleculeLoader
+        A MoleculeLoader object representing the sample FASTQ data.
+    """
+    from pyaptamer.data.loader import MoleculeLoader
+
+    fastq_path = os.path.join(os.path.dirname(__file__), "..", "data", "sample.fastq")
```
siddharth7113: Please make sure this is also covered in the `pyproject.toml` so we can ship the file with the installation.

pyaptamer/trafos/transform/_selex.py
```diff
@@ -0,0 +1,269 @@
...
+    def _check_X(self, X):  # noqa: N802
+        """Require a MoleculeLoader, then defer to the base coercion/checks."""
+        if not isinstance(X, MoleculeLoader):
+            raise TypeError(
+                f"{type(self).__name__} accepts only a MoleculeLoader as input, "
+                f"got {type(X).__name__}."
+            )
+        return super()._check_X(X)
+
+    def fit(self, X, y=None):
```
siddharth7113: this is wrong, please check how `Basetransform` work, you should not be overriding it and instead overriding the private method `_fit`

pyaptamer/trafos/transform/_selex.py
```diff
@@ -0,0 +1,269 @@
...
+        verbose: bool = False,
+    ):
+        self.method = method
+        self.max_len = max_len
+        self.start_primer = start_primer
+        self.end_primer = end_primer
+        self.tolerance = tolerance
+        self.verbose = verbose
+
+    def _check_X(self, X):  # noqa: N802
```
siddharth7113: I don't think we should force both here acceptring dataframe should be fine too.

pyaptamer/datasets/_loaders/_fastq_loader.py
```diff
@@ -0,0 +1,19 @@
...
+
+    return MoleculeLoader(data={"sequence": [fastq_path]}, tiling="samples")
```
siddharth7113: Can you check if tests are passing if you don't pass tiling=`samples` here? becasue I remeber this is not default here, and could cause trouble in transformer.

pyaptamer/trafos/transform/_selex.py
```diff
@@ -0,0 +1,269 @@
...
+
+import numpy as np
+import pandas as pd
+
+from pyaptamer import logger
+from pyaptamer.data import MoleculeLoader
+from pyaptamer.trafos.base import BaseTransform
+
+
+def _infer_adapter(
```
siddharth7113: We should skip this for now and should be part of a sepatate issue, I want to see we could find a more effiecent implementation for this, we do not need to follow aptadiff for it. Maybe put it as a todo comment in code and open an issue and reference that issue here.

siddharth7113:
Hi @aditi-dsi  I pushed changes straight to your branch, Most of this came out of testing edge cases, not from anything wrong with the approach. 


The sample file wasn't getting installed. It was missing from the package config.


Renamed the class to `PrimerTrimmer`.

Removed the heuristic primer guessing. When it couldn't find a primer it returned an empty string, and then every read counted as matching, so sequences came back untrimmed while reporting success. It could also cut the primer short, or guess primers longer than the reads themselves. None of those threw an error, so we just get wrong sequences. We can add heuristic later as separate PR and explore other approached as well.

We now pass the random region length instead of it being learned from the data. I removed tolerance as well, since we would only the fixed size random region.

Bad reads are dropped instead of converting it to NA. If every read gets dropped you now get a warning, since that means the primers are wrong.

A plain DataFrame works again, and passing a loader with the wrong tiling now tells you to use tiling="samples".

aditi-dsi:
> Bad reads are dropped instead of converting it to NA.

shouldn't the number of input samples and output samples be same as per scikit-learn convention for `transform`?

pyaptamer/trafos/tests/test_primer_trimmer.py
```diff
@@ -1,140 +1,155 @@
...
+    core = "ACGT" * 8
+    X = MoleculeLoader(data={"sequence": [make_read("AAA" + core + "TTT")]})
+
+    once = PrimerTrimmer(START_PRIMER, END_PRIMER, len(core) + 6).fit_transform(X)
+    twice = PrimerTrimmer("AAA", "TTT", len(core)).fit_transform(
+        MoleculeLoader(data=once)
+    )
+
+    assert once["sequence"].tolist() == ["AAA" + core + "TTT"]
+    assert twice["sequence"].tolist() == [core]
```
aditi-dsi: why is the need for feeding the already-trimmed region back for another trimming?

pyaptamer/trafos/transform/_primer_trimmer.py
```diff
@@ -0,0 +1,174 @@
...
+            If the first column of X does not hold str reads.
+        """
+        kind = pd.api.types.infer_dtype(X.iloc[:, 0], skipna=True)
+
+        if kind != "string":
+            raise TypeError(
+                f"{type(self).__name__} expects one str read per row, but the "
+                f"first column of X is {kind!r}. If it holds several reads per "
+                "cell, the MoleculeLoader was built with the default "
+                'tiling="bag" - use tiling="samples" so each read is one row.'
```
aditi-dsi, line 141: since `load_sample_fastq()` already builds with `tiling="samples"`. In which case this can occur?

pyaptamer/trafos/tests/test_primer_trimmer.py
siddharth7113: this was for adapters along with primer region , I would want it work for that case, but I think , it could be directly passed once as together so we could remove it.

pyaptamer/trafos/transform/_primer_trimmer.py
siddharth7113: In all other cases, where you are not using `load_sample_fastq()` since this would be used with real raw data and `MoleculeLoader` defaults to `bag` when you load a new data file.

siddharth7113 (dismissed):
Changes being worked on by @aditi-dsi for dropping or making sequences NaN

aditi-dsi:
@siddharth7113, the changes has been pushed, it's ready for re-review.

pyaptamer/trafos/transform/_primer_trimmer.py
```diff
@@ -0,0 +1,246 @@
...
+    start_primer : str
+        The constant 5' region, matched as a literal prefix of each read.
+    end_primer : str
+        The constant 3' region, matched as a literal suffix of each read,
+        given in the orientation in which it appears in the read.
+    variable_length : int
+        The designed length of the random region, for example 40 for an N40
+        library.
+    on_unmatched : {"drop", "na", "raise"}, default="drop"
+        What to do with a read that fails to match the design.
```
siddharth7113: What does this mean? Could you put a simpler sentence and a small nit maybe use one keyword arg here?

pyaptamer/trafos/transform/_primer_trimmer.py
```diff
@@ -0,0 +1,246 @@
...
+          ``None``. ``Xt`` holds one row per input read.
+        - "raise" : raise a ``ValueError`` if any read fails to match.
+
+    Notes
+    -----
+    Estimating the constant regions or the library design from the reads is
+    not supported. See the AptaDiff project for prior art on inferring them.
+
+    Examples
+    --------
```
siddharth7113: Could you add another example where this new variable is being used?

siddharth7113 (approved):
I have updated the docstrings and comment and a small test case fix, overall rest of the things looks good.

siddharth7113:
@aditi-dsi  I recently merged testing across all transformers, this might be cuasing failure in CI, Could you check and see what is causing these failures and if possible fix thoses here?

aditi-dsi:
@siddharth7113, the failures were due to missing `get_test_params` needed by the skbase tests after your recent merge. Added that in PrimerTrimmer and the CI tests are passing now. Also, as discussed, removed the redundant `test_output_can_be_fed_back_in` added by you earlier.

PR #761, opened 2026-08-18
Raptgen pipeline draft

### LINKED ISSUE #743: [ENH] RaptGen pipeline
Refs #709
Adding RaptGen training/usage pipeline class `RaptGenPipeline ` in `/raptgen/_pipeline.py`  following the sklearn style pattern used in aptanet

### Proposed API

RaptGenPipeline(BaseEstimator, TransformerMixin) with:
- __init__()
- fit(X, y=None): trains CNN_PHMM_VAE on aptamer sequences (unsupervised, y unused but kept for sklearn API compatibility)
- transform(X): runs sequences through the trained encoder, returns latent-space points
- inverse_transform(X): generates sequences from latent-space points, internally uses ProfileHMMSampler to produce state/emission probabilities, then the decoder to turn those into a final sequence

NoorMajdoub:
#### Reference Issues/PRs
Closes #743 (draft)
Depends on #741

RaptGenPipeline(BaseEstimator, TransformerMixin) wrapping CNN_PHMM_VAE:

- fit(X, y=None) : trains on aptamer sequences

- transform(X) : encodes sequences to latent points 

- inverse_transform(Z, most_likely=True) : generates sequences from latent points

Please just check pyaptamer/raptgen/_pipeline.py
**Two questions**
- Train/val split: raptgen original train() takes pre-split loaders as args. To fit sklearn's fit(X, y=None), I moved the split into a private _train_val_split helper called inside fit. Rest of the training loop is a direct port. 

- transform and inverse_transform : _build_model instantiates CNN_PHMM_VAE directly inside the function and transform and inverse_transform both reach into model_.encoder/model_.decoder, do I continue with this approach?

Wanted to verify the overall design before adding the tests and improving the documentation for this.


siddharth7113:
Hi @NoorMajdoub,

I have spent some time going through the algorithm architecture and on a high level it can be classified at 3 levels, I understand this is still draft but the following info would be helpful while writing the code.:

- **Network blocks** : Mostly `nn.Module`. It only does the forward pass and does not know how to train itself. For RaptGen this is `CNN_PHMM_VAE` and everything in `layers/`.

- **Complete engine** :  A estimator that takes numeric arrays. Its fit does the training: make tensors, make batches with a DataLoader, run epochs, keep the best weights. This is like `MLPClassifier` in sklearn, or `AptaNetClassifier` here. The `DataLoader` should be part of this.

- **End to End algorithm** : The would be the first class a user calls to run the algorithm. It takes a `MoleculeLoader`, uses a `transformer` to turn strings into numbers, and passes the numbers to the engine. It should  not training code of its own.

- ` _pipeline.py` is level 2. 
-  To fit the layering, the string to index encoding in _encode_sequences should move into a small BaseTransform (on a side note: could you also check there is nothing that fits this description already), so the engine takes an integer array. 
- The training loop should go through a `LightningModule` and `Trainer`.

I have also attached the image that would help in understanding these levels. I would recommend renaming this from `RaptgenPipeline` to either `RaptgenGenerator` or a better name that suits the context, the goal is to get rid of the pipeline word from all the algorithms.

image

siddharth7113:
cc @aditi-dsi Aptadiff should follow this pattern as well.

PR #762, opened 2026-08-19
[ENH] AptaDiff: Implements core model and training wrapper

### LINKED ISSUE #758: [ENH] AptaDiff: Implement denoiser and diffusion wrapper with training-loss
The next component in AptaDiff is to implement the model architecture and and orchestrate the training loss elements.

#### Scope of Work
- **Model:** Implement constructor to parse configuration parameters (e.g., diffusion steps, transformer dimensions) and instantiate the transformer-backed denoiser and diffusion wrapper.
- **Training Loss:** Wire the diffusion wrapper's `log_prob` path, timestep sampling, forward diffusion (q_sample), and the denoiser call, around the already-implemented VLB loss func. 
- **Validation:** Add validation pass and testing hooks to confirm the model can successfully compute a training loss on a small batch with correct tensor shapes and gradients.

#### Dependencies
- Builds on top of the transformer layer porting with core diffusion utils ([#739](https://github.com/gc-os-ai/pyaptamer/pull/739)).

aditi-dsi:
#### Reference Issues/PRs

closes #758 
stacks on #739

This PR introduces the `_model.py` module, implementing the `AptaDiffDenoiser` neural network backbone wrapper and the `AptaDiffDiffusion` orchestration wrapper for diffusion process, alongside its corresponding tests.

Yes, in `pyaptamer/aptadiff/tests/test_model.py`
#### PR checklist

- [x] Added/modified tests


siddharth7113:
Can you instead base this PR on #739 branch, I think github now supports stacked PR. this would easier to keep track with the diffusion algorithm.

aditi-dsi:
> Can you instead base this PR on #739 branch, I think github now supports stacked PR. this would easier to keep track with the diffusion algorithm.

yes, this PR is already based on PR 739.

siddharth7113:
> > Can you instead base this PR on #739 branch, I think github now supports stacked PR. this would easier to keep track with the diffusion algorithm.
> 
> yes, this PR is already based on PR 739.
image


this shows it's from main make the base branch not main instead branch out from 739

aditi-dsi:
> > > Can you instead base this PR on #739 branch, I think github now supports stacked PR. this would easier to keep track with the diffusion algorithm.
> > 
> > 
> > yes, this PR is already based on PR 739.
> 
> image
> this shows it's from main make the base branch not main instead branch out from 739

Oh I see what you mean now. For this, the base branch has to live in gc-os-ai for it to show up in the dropdown, and the #739 branch is only on my fork. Tried pushing it up and got a 403.

siddharth7113:
> > > > Can you instead base this PR on #739 branch, I think github now supports stacked PR. this would easier to keep track with the diffusion algorithm.
> > > 
> > > 
> > > yes, this PR is already based on PR 739.
> > 
> > 
> > image
> > this shows it's from main make the base branch not main instead branch out from 739
> 
> Oh I see what you mean now. For this, the base branch has to live in gc-os-ai for it to show up in the dropdown, and the #739 branch is only on my fork. Tried pushing it up and got a 403.

Okay, makes sense

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+
+def _log_onehot_to_index(log_x: torch.Tensor) -> torch.Tensor:
+    """Recover integer class indices from a log-one-hot tensor.
+
+    Parameters
+    ----------
+    log_x : torch.Tensor
+        Log-probabilities of shape (batch_size, num_classes, seq_len), as
+        produced by :func:`_index_to_log_onehot` or the forward diffusion
+        process.
```
siddharth7113, line 33: I understand that the `log_x` is produced from the `_index_to_log_onehot` but I didn't understand "or the forward diffusion process" do you mean? there is a separate way that could produce it too? then it would be good to emntion the exact class or function that could do that or is it just a small typo and should be "in the forward diffusion process"?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+
+    Returns
+    -------
+    torch.Tensor
+        Integer class indices of shape (batch_size, seq_len), obtained by
```
siddharth7113, line 38: What is Integer class here?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+        taking the argmax over the class dimension.
+    """
+    return log_x.argmax(dim=1)
```
siddharth7113, line 41: nit: Do you think we need this fucntion, if it is being repeated constantly then I would be okay but at this point, would it not be better to use at the point where it needs and mention the shape change there, I could see no other benifiti otherwise.

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+
+
+def _log_sample_categorical(logits: torch.Tensor, num_classes: int) -> torch.Tensor:
+    """Sample class indices from unnormalized logits via the Gumbel-max trick.
```
siddharth7113, line 45: Add a reference to gumbel noise please, I have no clue what is gumbel noise and no way from here to reference it.

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+
+    x_onehot = F.one_hot(x, num_classes)
+    x_onehot = x_onehot.permute(0, 2, 1).float()
+    eps = torch.finfo(x_onehot.dtype).tiny
+
+    return torch.log(x_onehot.clamp(min=eps))
+
+
+class Rezero(nn.Module):
+    """Learnable scalar gate, initialized to zero, that scales its input.
```
siddharth7113, line 112: Docstrings can be improved, I am not sure what is learnable scalar gate?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+
+    `alpha` starts at zero, so the module's output is exactly zero for any
+    input at initialization, and grows away from zero as `alpha` is
```
siddharth7113, line 115: I am unclear what does grow away from zero means? is it going upwards or downwards or both ways?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+        ----------
+        x : torch.Tensor
+            Input tensor of any shape.
+
+        Returns
+        -------
+        torch.Tensor
+            `x` scaled elementwise by `alpha`, same shape as `x`.
+        """
+        return self.alpha * x
```
siddharth7113, line 142: again same thought why does this need to be a sepatate class and not part of the same AptaDiffDenoiser and AptaDiffDiffusion class, directly as an arg.

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+    denoise_fn : nn.Module
+        The denoising network, e.g. an `AptaDiffDenoiser` instance. Must
+        accept `(x, t, z)` and return logits of shape
+        `(batch_size, num_classes, seq_len)`.
+    num_classes : int, optional, default=4
+        The number of unique nucleotides in the sequence.
+    num_timesteps : int, optional, default=1000
+        Total number of diffusion timesteps.
+    loss_type : {"vb_stochastic", "vb_all"}, optional, default="vb_stochastic"
+        Which variational bound to optimize.
```
siddharth7113, line 283: small nit: "chooses which variational bound to optimise"

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+        alphas = cosine_alpha_schedule(num_timesteps)
+        log_alpha = torch.log(alphas)
+        log_alphabar = torch.cumsum(log_alpha, dim=0)
+
+        log_1m_alpha = log_one_minus_exp(log_alpha)
+        log_1m_alphabar = log_one_minus_exp(log_alphabar)
+
+        self.register_buffer("log_alpha", log_alpha)
+        self.register_buffer("log_1m_alpha", log_1m_alpha)
+        self.register_buffer("log_alphabar", log_alphabar)
```
siddharth7113, line 377: What is the use for these buffer? On a quick look I couldn't find the reference for it?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,753 @@
...
+                log_x0=log_x0,
+                log_xt=log_xt,
+                t=sampled_timesteps,
+                log_alpha=self.log_alpha,
+                log_alphabar=self.log_alphabar,
+                log_pred=log_pred,
+            )
+
+            kl_prior = self.kl_prior(log_x0)
+            total_loss = (loss / sampled_probs) + kl_prior
```
siddharth7113, line 751: A lot of code feels repeated from _train_loss, maybe not 100% but could you make sure they are abstracted at one place only.

siddharth7113 (changes requested):
Hi Aditi,

The PR is looking good, I have done a first review of _model.py and left some feedback on mostly docstrings, nits and some chuncks of repeated code, I have yet to do through review of some parts such as tests, and some functions in `AptaDiffDiffusion` class but for the meanwhile you can make the mentioned changes.

Also a small usage example for `AptaDiffDiffusion` would be nice, doesn't need to be too big a simple implementation of implementing and calling loss should be fine.

pyaptamer/aptadiff/_model.py
aditi-dsi: Yes, specifically it can be produced by `q_sample`, I'll add that note in docstring.

pyaptamer/aptadiff/_model.py
aditi-dsi: "Integer class indices" means discrete categorical token IDs (0..K-1 for sequence chars) represented as integer tensors.

Will improve the wording.

pyaptamer/aptadiff/_model.py
aditi-dsi: While it only has one call site in this specific Level-1 file, this conversion will be needed repeatedly by other methods in Level-2 generation engine to turn model outputs back into token indices during sampling loops.

pyaptamer/aptadiff/_model.py
aditi-dsi: The transformer passes its output through this gate/layer to scale its output by an `alpha`, which starts at 0 and is updated during training. 

Will improve the wording.

pyaptamer/aptadiff/_model.py
aditi-dsi: It means both ways, `alpha` is an unconstrained param, so optimizer can update it in either direction (up or down) away from 0.

pyaptamer/aptadiff/_model.py
aditi-dsi: I kept is as a self-contained class to keep the two learning processes (transformer and rezero) separate and readable. But, since it's used only inside denoiser, maybe I can move it there itself.

pyaptamer/aptadiff/_model.py
aditi-dsi: Good catch. You're partially right here, `log_1m_alpha` and `log_1m_alphabar` is dead code left while porting from upstream ref, so I'll remove it. But the other buffers are actively used for the discrete forward diffusion process: 
- `log_alpha` is used in `q_pred` to compute the single-step transition proba
- log_alphabar: Used in `q_posterior` and `_train_loss` to compute the marginal distributions.

They are registered as buffers so these non-trainable constants automatically transfer to the correct GPU/CPU device during `.to(device)` calls and persist in the model's `state_dict` without tracking gradients.

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,683 @@
...
+    ...     output_dim=4,
+    ...     dim=32,
+    ...     depth=1,
+    ...     n_blocks=1,
+    ...     max_seq_len=8,
+    ...     num_timesteps=50,
+    ...     heads=2,
+    ...     local_attn_window_size=8,
+    ... )
+    >>> diffusion = AptaDiffDiffusion(denoise_fn=denoiser, num_timesteps=50)
```
siddharth7113, line 281: What would happen if I pass different `num_timesteps` to both? Do they need to be same?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,683 @@
...
+           https://arxiv.org/abs/2102.09672
+
+    Examples
+    --------
+    >>> import torch
+    >>> import torch.nn.functional as F
+    >>> from pyaptamer.aptadiff import AptaDiffDenoiser, AptaDiffDiffusion
+    >>> denoiser = AptaDiffDenoiser(
+    ...     enc_embed_size=16,
+    ...     input_dim=4,
```
siddharth7113: I think this also be need to checked again `num_classes` for denoiser becuase they should be correlated right? Simple checks in `post_init` or `init` should resolve this.

pyaptamer/aptadiff/_model.py
siddharth7113: also a small nit: I think a boundary value test would be nice like running at a small timestep like `1` or whatever is min would ensure everything loads correctly.

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,683 @@
...
+                f"{z.size(0)}."
+            )
+
+        if x.dim() != 3 or x.size(1) != self.num_classes:
+            raise ValueError(
+                f"x must be one-hot of shape (batch_size, {self.num_classes}, "
+                f"seq_len), got {tuple(x.shape)}."
+            )
+
+        x = x.float().transpose(1, 2).contiguous().transpose(1, 2)
```
siddharth7113, line 706: Why is `transpose` here called twice?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,683 @@
...
+        log_x0 = x.clamp(min=torch.finfo(torch.float32).tiny).log()
+
+        if self.training and self.loss_type == "vb_all":
+            return -self.compute_full_vlb(log_x0, z)
+
+        return -self._stochastic_vlb(log_x0, z, update_statistics=self.training)
```
siddharth7113, line 712: Shouldn't we use `vb_all` for inference and `stochastic_vlb` for training, can you verify this is how it's done in original implementation, If so I would suggest to expose this as an option to user, so they could choose to use either one.

Note: My current limited understanding is the `vb_all` version is expensive , right?

pyaptamer/aptadiff/_model.py
```diff
@@ -0,0 +1,683 @@
...
+        Returns
+        -------
+        torch.Tensor
+            Log-probabilities for the model's reverse-step prediction,
+            shape (batch_size, num_classes, seq_len).
+            If `self.parametrization` is `"x0"`, this is the true posterior
+            q(x_{t-1} | x_t, predicted x0)
+            If `"direct"`, this is the denoiser's raw prediction.
+        """
+        if torch.any((t < 0) | (t >= self.num_timesteps)):
```
siddharth7113: nit: Could you modify this to `if ((t < 0) | (t >= self.num_timesteps)).any():` it's easier to read.

siddharth7113 (changes requested):
Hey Aditi,

Most of it looks good, I have left some comments about validation checks and a minor question about loss/scoring mechanism

pyaptamer/aptadiff/_model.py
aditi-dsi: The first transpose (BVL -> BLV) allows `.contiguous()` to make data sequential in-memory first, and the second transpose simply flips the shape back to what the model expects (BLV -> BVL).

pyaptamer/aptadiff/_model.py
aditi-dsi: > What would happen if I pass different `num_timesteps` to both? Do they need to be same?

Yes, correct. Different values will silently break the logic and and both classes will diverge. But I think that will be avoidable because of our 3-layer arch. User will have the option to pass only a single `num_timesteps` arg through the public interface (layer 3 pipeline) which will be drilled down to these two classes via layer-2 generator/estimator, which therefore will pass that single shared arg to both these classes while instantiation.

pyaptamer/aptadiff/_model.py
aditi-dsi: see [comment](https://github.com/gc-os-ai/pyaptamer/pull/762#discussion_r4069890683)

aditi-dsi: I'll rather add a note in docstrings for both these args, that they must match between both these classes, just in case if someone experiments with these independently and not via pipeline.

pyaptamer/aptadiff/_model.py
aditi-dsi: Yes, `vb_all` is expensive. The upstream code uses the stochastic estimate by default for training, and `vb_all` only when the user explicitly chooses to do so (with the authors' recommendation to avoid it). For eval, it strictly uses the stochastic one to keep evaluation rounds fast on validation splits.

And the same approach is followed for pyaptamer implementation.

siddharth7113: > The upstream code uses the stochastic estimate by default for training, and vb_all only when the user explicitly chooses to do so (with the authors' recommendation to avoid it).

How would they do it, which arg exposes it?

pyaptamer/aptadiff/_model.py
siddharth7113: > Yes, correct. Different values will silently break the logic and both classes will diverge. But I think that will be avoidable because of our 3-layer arch.

The pipeline will be fine, but these two classes are public and the docstring example uses them directly, which can be a problem when an individual tries to build using them with different values.
what I would do instead -
Since AptaDiffDiffusion already holds the denoiser, it should not need `num_timesteps` and `num_classes` a second time. Either-
1. Store `self.num_timesteps` and `self.num_classes` on AptaDiffDenoiser, and AptaDiffDiffusion reads them from `denoise_fn`. Then there is no scope of being wrong, and it would remove the ambiguity.
or
2. Keep the args on AptaDiffDiffusion, but in `__init__` compare them against `denoise_fn` and raise ValueError if they differ. Use `getattr` with a default so a custom denoise_fn without those attributes still works.

I would go with 2, since the tests already pass a plain nn.Module as denoise_fn.

siddharth7113 (commented):
Left some more comments

pyaptamer/aptadiff/_model.py
aditi-dsi: Via `loss_type` arg in AptaDiffDiffusion

pyaptamer/aptadiff/_model.py
aditi-dsi: Makes sense. Though I'd need to combine elements from both options, 1. storing the attrs and then 2. will use getattr with a default to compare.

aditi-dsi:
I've addressed the new comments and added the validation check for both the common args of the two classes as discussed.

pyaptamer/aptadiff/_model.py
siddharth7113: but I am assuming that doesn;t allow to change for inference right? suppose I would want to use the `vb_all` for inference, we can mention in docstring it differs from original implementation otherwise I thjink it would be give to user choice and keep the original as default.

aditi-dsi: Yes, choice during inference is not supported in current behaviour. So from what I understand, we want to drop the `self.training` branching and give both modes the same behaviour, i.e., `vb_stochastic` as default, and `vb_all` as another choice?

siddharth7113: No there is no pointing in droping behaviour , instead provide both options at both stages but put the defaults as whatever aptadiff paper has

siddharth7113 (approved):
Looks good , Thanks !

PR #766, opened 2026-08-27
[ENH] MoleculeLoader.iter_dataframe for chunked loading of large files

### LINKED ISSUE #765: [ENH] chunked loading in MoleculeLoader for large FASTQ files
## Problem

`MoleculeLoader` does not read files until `to_dataframe()` is called, but then it reads everything at once:

- `_load_seqio` calls `list(SeqIO.parse(...))`, so the whole file is loaded into memory as `SeqRecord` objects. These take several times the on-disk size.
- The result is a single pandas DataFrame holding all rows.

SELEX FASTQ files can be several GiB, so loading one this way can run out of memory.

Note that streaming the file read alone is not enough: with `tiling="samples"`, a 10 GiB FASTQ still becomes a ~10 GiB DataFrame. To really fix this, the caller must be able to receive the data in parts.

## Proposal

1. **Stream inside the loader (no API change).** Iterate over `SeqIO.parse` instead of building a list, and build rows as we go. This lowers peak memory, since we no longer hold the record list and the DataFrame at the same time. PDB files can stay as they are, since they are small.

2. **Add a chunked method.** A new method `iter_dataframe(chunksize=...)` that yields DataFrames of at most `chunksize` rows each. `to_dataframe()` stays unchanged. A separate method seems better than a `chunksize` argument on `to_dataframe`, because an argument that changes the return type is confusing. pandas moved away from this pattern for `read_csv`.

   Not every tiling can be chunked. `tiling="samples"` with one file column works row by row (the FASTQ case), and `bag`/`concat`/`first` work too. `features` and multi-column `samples` need to see the whole table first, so `iter_dataframe` would raise a clear error for those.

3. **Follow-up:** the chunk iterator is a building block for the sliceable loader in #706, which needs to read single rows or slices by position without loading the whole file.

## Why not dask

I looked at returning a `dask.DataFrame` and would not make it the base of the design:

- A dask DataFrame built from delayed file reads has one partition per file. A single big FASTQ is then still one big partition, so the within-file chunk reader has to be written either way.
- Several tiling options need to see the whole table (for example, `features` needs the largest sequence count over all rows), and the loader uses list-valued cells and custom indexes. dask handles these badly.
- Everything downstream in pyaptamer consumes pandas, so returning a dask object only moves the memory problem to the caller.

Once `iter_dataframe` exists, a `to_dask()` wrapper on top of it is a few lines, and dask can be an optional dependency for users who want out-of-core computation.

siddharth7113:
Fixes #765.

Adds `MoleculeLoader.iter_dataframe(chunksize)`, which yields DataFrame chunks so GiB-scale FASTQ files can be processed without loading the whole table. Internally, files are now streamed record by record instead of read with `list(SeqIO.parse(...))`. `to_dataframe()` is unchanged, and concatenating all chunks gives the same table. `tiling="features"` and `multiindex="auto"` raise in `iter_dataframe`, since they need the whole table before the first chunk.

Stacked on #752 to use its `sample.fastq` / `load_sample_fastq()` in tests; will be rebased onto `main` once #752 merges. Only the last commit is new.

### DIFF
--- a/pyaptamer/data/loader.py
+++ b/pyaptamer/data/loader.py
@@ -16,7 +16,8 @@ class MoleculeLoader:
 
     Holds tabular data whose cells may be file references (FASTA, PDB,
     FASTQ, ...) or in-memory values (sequence strings, numbers). Files are
-    not parsed until ``to_dataframe`` is called.
+    not parsed until ``to_dataframe`` or ``iter_dataframe`` is called. For
+    files too large to hold in memory at once, use ``iter_dataframe``.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


siddharth7113:
Superseded by #773, which makes the internals stream first. Chunked output (iter_dataframe) will be revisited on top of that, tracked in #765.

PR #773, opened 2026-09-07
[ENH] MoleculeLoader: stream file records, fast FASTQ reader, gzip support

siddharth7113:
Part of #765

## What changed

- Files are read record by record. `_cell_records` used to get a DataFrame
  built from `list(SeqIO.parse(...))`, which held every `SeqRecord` in
  memory at once and then converted again from DataFrame into tuples. Now a
  reader yields `(chain_id, sequence)` tuples and `_cell_records` collects them at once.

- One reader per format behind a dispatch dict in `_iter_records`:
  `_read_pdb` (SEQRES, chain id split), `_read_fastq` (Biopython's
  `FastqGeneralIterator`, plain strings, no `SeqRecord`), and `_read_seqio`
  as the fallback for every other format `Bio.SeqIO` knows. Adding a reader
  is one dict entry and one method.
- `.gz` files are read directly. `_determine_type` looks past a trailing
  `.gz` and `_open_text` opens with `gzip.open(path, "rt")`. Files from ENA
  and SRA arrive gzipped, and until now they failed with
  `Unknown format 'gz'`.
- `_tile_samples` collects one list per column and builds the DataFrame
  once, instead of one dict per output row.

## Measured

BARX1 HT-SELEX cycle 1 from Jolma et al. 2013 (ENA ERR188678, 17.7 MB,
216,997 reads), `tiling="samples"`, timed with `time.perf_counter`:

https://www.ebi.ac.uk/ena/browser/view/ERR188678

| | before | after |
|---|---|---|
| `to_dataframe()` | 32.5 s | 0.82 s |
| memory growth during the call | ~290 MB | ~105 MB |

## Tests

`test_loader.py` now has one test per reader and tiling path.
New: gzipped FASTA and FASTQ are read; `_determine_type` is checked
directly on six filenames (plain, `.gz`, uppercase, extra dot, `.gz`
only, no suffix); FASTQ is read by `FastqGeneralIterator` and not parsed through
`SeqIO.parse`; In FASTA
chain id should be the first word of the header, as in `SeqRecord.id`; a FASTQ
whose quality lines start with `@` is parsed correctly; a `samples` row
with no file cell gives one row; `features` keeps literal and
single-sequence columns by name.

### DIFF
--- a/pyaptamer/data/loader.py
+++ b/pyaptamer/data/loader.py
@@ -3,12 +3,14 @@
 __all__ = ["MoleculeLoader"]
 __author__ = ["fkiraly", "satvshr", "siddharth7113"]
 
+import gzip
 import os
 from itertools import product
 from pathlib import Path
 
 import pandas as pd
 from Bio import SeqIO
+from Bio.SeqIO.QualityIO import FastqGeneralIterator
 
 
 class MoleculeLoader:
@@ -45,6 +47,8 @@ class MoleculeLoader:
     Supported file formats are anything :func:`Bio.SeqIO.parse` can read
     (FASTA, GenBank, EMBL, FASTQ, ...); PDB files are read through the
     ``pdb-seqres`` parser. The format is inferred from the file suffix.
+    Gzipped files are read directly: ``reads.fastq.gz`` is decompressed
+    and parsed as FASTQ.
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
I think this is a great refactor.

As far as I understand, the key change is of `_load_dispatch` from a single-return to a yield architecture, which makes sense.

I would still though:

* ensure the docstrings stay complete, or this will be hard to read. Parameters and returns should always be documented, as should behaviour, assumes and guarantees. Eroding this in a key API location is a bad idea...
* keep the explicit dispatch architecture intact, and very explicit, just do it on the level of chunks as you are proposing

fkiraly (approved):
Looks great, thanks!

PR #776, opened 2026-09-08
[BUG] BaseTransform: univariate transformers select their column by p…

### LINKED ISSUE #770: [BUG] BaseTransform test hardcodes "seq" column name, restricting named-column transformers
**Describe the bug**

The current transformations test suite hardcodes the column name it feeds to transformers. `_SequenceFrameScenario` in `pyaptamer/trafos/base/tests/test_base.py` builds `pd.DataFrame({"seq": [...]})` for both fit and transform, and it applies to every transformer tagged `capability:multivariate: False`.

That works fine for `PrimerTrimmer` and `GreedyEncoder` because they pick their column positionally (`X.columns[0]` and `X.values[:, 0]`). Any transformer that selects its column by name has to be named `"seq"` or it fails with a `KeyError`, so the test fixture ends up controlling what the estimator's parameter default should be.

**To Reproduce**

Any transformer that picks its column by name hits this. With a `sequence_col`
default of anything other than `"seq"`, and no `get_test_params()` override:

```python
class DemoSequenceEncoder(BaseTransform):
    _tags = {"property:fit_is_empty": True, "capability:multivariate": False}

    def __init__(self, sequence_col="sequence"):
        self.sequence_col = sequence_col
        super().__init__()

    def _transform(self, X):
        if self.sequence_col not in X.columns:
            raise KeyError(
                f"{type(self).__name__} expects a column named "
                f"{self.sequence_col!r}, but X has columns {list(X.columns)}."
            )
        return X[[self.sequence_col]]
```

```
self = DemoSequenceEncoder(), X = seq
0  ACGU
1  GUAC

E   KeyError: "DemoSequenceEncoder expects a column named 'sequence',
    but X has columns ['seq']."

FAILED TestAllTransformers::test_fit_transform_sets_is_fitted[DemoSequenceEncoder]
```

And if we make `sequence_col` mandatory, it just moves the failure to `skbase._exceptions.FixtureGenerationError` , with a message that does not mention columns at all:

```
E   ValueError: skbase object <class '...SequenceKHotEncoder'> has parameters
    without default values, but these are not set in get_test_params.
```

**Expected behavior**

A transformer should be able to pick its column by name without the test suite controlling what that name has to be. The suite should either ask the transformer what column it wants, or build the fixture to match.

**Additional context**

Currently, [`SequenceKHotEncoder`](https://github.com/gc-os-ai/pyaptamer/pull/732) has a workaround for this by overriding `get_test_params()` to return `{"sequence_col": "seq"}`. That works and it is the normal skbase way to configure test instances. If someone changes `"seq"` in `_SequenceFrameScenario`, every name-based transformer like this would break.

Even after the fix, the `get_test_params()` method can stay, but it stops being load-bearing.

**Versions**
<details>

0.1.0a1
</details>

siddharth7113:
#### Reference Issues/PRs

Fixes #770.

The transformer test suite feeds univariate transformers ,a frame with a
column named `seq`. A transformer that selects its column by name fails
unless that name is `seq`, so the fixture was rigid.

The base class already rejects more than one column for univariate
transformers, so a column name parameter there can only match or raise.

This PR states the rule in the `_fit` and `_transform` docstrings: X has
one column, select it by position, its name is not fixed.

- The wording of the X parameter in `_fit` and `_transform`.

Yes. `test_univariate_accepts_any_column_name` renames the column and runs
fit then transform on every univariate transformer.

#### PR checklist
- [x] The PR title starts with [BUG].
- [x] Added/modified tests
- [x] Used pre-commit hooks when committing.

### DIFF
--- a/pyaptamer/trafos/base/_base.py
+++ b/pyaptamer/trafos/base/_base.py
@@ -53,8 +53,10 @@ class BaseTransform(BaseEstimator):
 
         Parameters
         ----------
-        X : array-like, shape (n_samples, n_features)
-            Input data to fit the transformer.
+        X : pd.DataFrame, shape (n_samples, n_features)
+            Input data to fit the transformer. If the tag
+            ``capability:multivariate`` is False, X has exactly one column.
+            Select it by position. Its name is not fixed.
         y : array-like, shape (n_samples,), optional
             Target values. Only used if the transformer has
             the tag ``capability:y`` set to True.
@@ -96,8 +98,10 @@ class BaseTransform(BaseEstimator):
 
         Parameters
         ----------
-        X : array-like, shape (n_samples, n_features)
-            Input data to transform.
+        X : pd.DataFrame, shape (n_samples, n_features)
+            Input data to transform. If the tag ``capability:multivariate``
+            is False, X has exactly one column. Select it by position. Its
+            name is not fixed.
 
         Returns
         -------
--- a/pyaptamer/trafos/base/tests/test_base.py
+++ b/pyaptamer/trafos/base/tests/test_base.py
@@ -143,6 +143,15 @@ class TestAllTransformers(TransformerFixtureGenerator, TestAllObjects):
         object_instance.fit_transform(**scenario.args["fit"])
         assert object_instance.is_fitted is True
 
+    def test_univariate_accepts_any_column_name(self, object_instance):
+        """A univariate transformer works on its one column whatever it is called."""
+        if object_instance.get_tag("capability:multivariate", False):
+            pytest.skip("only univariate transformers receive a single column")
+        args = _scenario_for(object_instance).args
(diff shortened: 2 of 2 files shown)


aditi-dsi:
@siddharth7113, LGTM. 

OOS: I noticed we don't have tests covering the case when`"capability:y"` is True (supervised transformers). Though, I believe we don't have any such transformer yet so it's non-blocking, but we can probably open a follow-up issue to add a supervised scenario later.

siddharth7113:
> @siddharth7113, LGTM.
> 
> OOS: I noticed we don't have tests covering the case when`"capability:y"` is True (supervised transformers). Though, I believe we don't have any such transformer yet so it's non-blocking, but we can probably open a follow-up issue to add a supervised scenario later.

I can not see a possibility of any transformer but sure if we decide to add one, adding a capability then would be fine.

